PR #23, opened 2023-01-29
Add Relief Staff and Regular Staff as Roles

### LINKED ISSUE #16: Configure 3 Authorization Levels
## Summary
- We require 3 different user roles: Admin, Regular Staff, and Relief Staff
- New user roles must be assigned on login and should be changeable
- 3 Point estimation
## Goals
- [ ] Create 3 new roles in the DB
- [ ] Connect to authentication so role is assigned on login
- [ ] Create REST endpoint to change a user's role (only admins should be authorized to use this endpoint)
## Implementation
- Start with adding the roles in the db (probably have to re-seed the db)
- Configure firebase auth stuff to handle and correctly assign the new roles
- Create a SQL query to update the user role (check if one exists already)
- Create a REST endpoint that calls the query and returns whether the change was successful or not
## Notes
- Probably a more difficult task, could be a good one to pair on

safchow:
## GitHub ticket link

[Configure 3 Authorization Levels](https://github.com/uwblueprint/supportive-housing/issues/16)

## Implementation description
* Added two new roles, Relief Staff and Regular Staff, to role definition. 
* On user creation, the user's role will be set to Relief Staff.
* The User role has been removed. The Admin role is unchanged.

## Steps to test
**0. Run Migrations**
```
$ docker exec -it SHOW-backend /bin/bash
$ flask db upgrade
```

**1. Front-end Test**
- Create a new user through the front-end.
- Verify that the db is updated with a new user, and the role is set to "Relief Staff".

**2. Back-end Test**
- On Postman, create a new request to the /auth/register endpoint.
- Use the following body: 
```
{
            "first_name": "Test",
            "last_name": "Script",
            "email": "testUserRole@test.com",
            "password": "password123"
}
```
- Grab a new auth token for the authorization of the request, you can do this by running the e2e tests and grabbing it from the output.
- Once the request is sent, ensure you get an ok response and the table is updated with the new user, and the role is set to "Relief Staff".

**3. e2e Tests**
- run all e2e tests and verify that they pass.

**4. Upgrade Role**
- Verify that a user can be upgraded to the "Regular Staff" Role.
- Verify that a user can be upgraded to the "Admin" role.

* 

## Checklist


safchow:
Todo:
- Create/change existing e2e tests to verify that when a user is created, the default role is Relief Staff

Things to look into:
- A possible db rebuild may be required for this change to take effect. Otherwise, requests will complain about invalid role.

DinuWije (changes requested):
Code for the 3 user roles look good. Can you also:
1. write a quick test to demo the ability to switch user roles via the API (i think the endpoint u would use is `PUT /users/id`)
3. update calls to `@require_authorization_by_role` with the new user role types

looks good otherwise!

backend/python/migrations/versions/dedc5f7be10c_initial_migration.py
```diff
@@ -36,7 +36,11 @@ def upgrade():
         sa.Column("first_name", sa.String(), nullable=False),
         sa.Column("last_name", sa.String(), nullable=False),
         sa.Column("auth_id", sa.String(), nullable=False),
-        sa.Column("role", sa.Enum("User", "Admin", name="roles"), nullable=True),
+        sa.Column(
```
connor-bechthold, line 39: From what I remember, we shouldn't be editing an existing migration but instead creating a new migration where the name of the migration is the change that we made. So, we'd want to have a migration file where we're adding relief staff as a role. Outlined [here](https://uwblueprint.github.io/starter-code-v2/docs/migrations/python-migrations.html#how-to-create-and-run-a-migration)

I may be incorrect with my thinking here, so this is definitely up to discussion.

backend/python/migrations/versions/dedc5f7be10c_initial_migration.py
```diff
@@ -36,7 +36,11 @@ def upgrade():
...
+            "role",
+            sa.Enum("Admin", "Regular Staff", "Relief Staff", name="roles"),
+            nullable=True,
```
connor-bechthold, line 42: Do we want this to be nullable? Or should we require every user we create has a role

connor-bechthold (commented):
Only general question I have is if we want to explicitly define the roles column to be `nullable = False` in the User table

backend/python/migrations/versions/dedc5f7be10c_initial_migration.py
safchow: you're right, a new migration file was created for the changes and the initial migration has been restored

backend/python/migrations/versions/dedc5f7be10c_initial_migration.py
safchow: should be false, updated

backend/python/app/rest/user_routes.py
```diff
@@ -113,7 +113,7 @@ def create_user():
 
 
 @blueprint.route("/<int:user_id>", methods=["PUT"], strict_slashes=False)
```
DinuWije, line 115: can we restrict `create_user()` and `delete_user()` endpoints to `Admin`? This may or may not break some tests, not sure.

backend/python/migrations/versions/dedc5f7be10c_initial_migration.py
DinuWije: I think u might have forgotten to push the change for this.

DinuWije (changes requested):

backend/python/migrations/versions/dedc5f7be10c_initial_migration.py
safchow: this is for the initial migration, on the new migration it should be false

backend/python/app/rest/user_routes.py
safchow: what about the `get_users()` and `update_user()` endpoints? I think the `update_user()` should be admin only, since its updating the users role but not sure about `get_users()`

DinuWije: ur right but i think that's out of scope for this pr

backend/python/migrations/versions/dedc5f7be10c_initial_migration.py
DinuWije: u right

DinuWije (commented):
LGTM demon 🚀

DinuWije (approved):

connor-bechthold (approved):
Congratulations

Issue #20, opened 2023-01-28
Implement logging of sign-on data


DinuWije:
## Summary
- The SHOW admin want to be able to see logs of who has signed in and when
- 3 point estimation
## Goals
- [ ] Use the information for the sign-on endpoints to curate a log of sign-on data
- [ ] Create an API endpoint to retrieve the logs. Optional query params/json body should be log start/end dates, and user email for the logs
## Implementation
- Either store the logging data in an existing table (probably not the best idea) or create a new table to store logs
- Write a SQL query to retrieve the logs from the db
- Create an REST GET endpoint to call the SQL query (remember to use the optional params)
## Notes
- Pretty good starter backend task

PR #25, opened 2023-02-09
Sign In Logs Issue #20

phamkelly17:
## ticket link
https://github.com/uwblueprint/supportive-housing/issues/20

## Implementation description
* Created the sign in log table in db
* add to sign in log table after each login
* Create API endpoint to search the logs by email, date range, or both

## Steps to test
1. Test if the response is correct
2. Test if the user is added to the sign in log table after signing in

* for the API endpoint, the start_date and end_date are passed in as strings, should they be passed in as type datetime? If so, how do we send datetime variables in Postman?
* The response for the API endpoint returns the logs with the most recent at the top, is this OK?

## Checklist
- [ x] My PR name is descriptive and in imperative tense
- [ x] My commit messages are descriptive and in imperative tense. My commits are atomic and trivial commits are squashed or fixup'd into non-trivial commits
- [x ] I have run the appropriate linter(s)
- [ x] I have requested a review from the PL, as well as other devs who have background knowledge on this PR or who will be building on top of this PR
- [ x] If I have made API changes, I have updated the [REST API Docs https://www.notion.so/uwblueprintexecs/Sign-In-Logs-Endpoint-Documentation-e0cb1c3804d84ee381775219df3a8ac9?pvs=4 
- [x ] IF I have made changes to the db/models, I have updated the [Data Models Page] https://www.notion.so/uwblueprintexecs/Data-Models-760f8aa06b244eb0842c079ad77987b0
- [ x] I have updated other Docs as needed


helenellyx (approved):
lgtm!

backend/python/app/rest/logs_routes.py
```diff
@@ -0,0 +1,97 @@
...
+
+
+@blueprint.route("/", methods=["GET"], strict_slashes=False)
+@require_authorization_by_role({"Admin"})
+def filter_logs():
+    email = request.json["email"]
+    start_date = request.json["start_date"]
+    end_date = request.json["end_date"]
+
+    if email and (start_date and end_date):
```
DinuWije: You don't need brackets here. Not an issue but just a code cleanliness thing.

backend/python/app/rest/logs_routes.py
```diff
@@ -0,0 +1,97 @@
...
+            return (
+                jsonify(
+                    {
+                        "error": "start_date and end_date query parameter must be datetime and email must be a string"
+                    }
+                ),
+                400,
+            )
+        elif start_date > end_date:
+            return jsonify({"error": "start_date must be before end_dateg"}), 400
```
DinuWije:
```suggestion
            return jsonify({"error": "start_date must be before end_date"}), 400
```

backend/python/app/rest/logs_routes.py
```diff
@@ -0,0 +1,97 @@
...
+            )
+            logs_list = sign_in_logs_service.get_logs_list(logs)
+            return logs_list, 200
+        except Exception as e:
+            error_message = getattr(e, "message", None)
+            return (
+                jsonify({"error": (error_message if error_message else str(e))}),
+                500,
+            )
+    elif email and not (start_date or end_date):
```
DinuWije: This should be 
```suggestion
    elif email and not (start_date and end_date):
```
thoughts?

backend/python/app/rest/logs_routes.py
```diff
@@ -0,0 +1,97 @@
...
+        if type(email) is not str:
+            return jsonify({"error": "email query parameter must be a string"}), 400
```
DinuWije, line 36: These conditionals are a bit confusing. Could you refactor them so that we have:
- One check for valid email in the beginning
- One check inside the first if statement for whether the start and end dates are valid (string with start < end)

backend/python/app/rest/logs_routes.py
```diff
@@ -0,0 +1,97 @@
...
+        else:
+            try:
+                user_id = user_service.get_user_by_email(email).id
```
DinuWije, line 39: Also here, can you try refactor some logic out so that you don't have repeated code?

backend/python/app/rest/logs_routes.py
```diff
@@ -0,0 +1,97 @@
...
+                        "error": "start_date and end_date query parameter must be datetime"
+                    }
+                ),
+                400,
+            )
+        elif start_date > end_date:
+            return jsonify({"error": "start_date must be before end_dateg"}), 400
+        try:
+            # return as json object
+            logs = sign_in_logs_service.get_logs_by_date_range(start_date, end_date)
```
DinuWije, line 86: Not saying you need to change this, but I'm wondering why you return the logs in the `get_logs_by_date_range` method and then pass them back into the `logs_list` method? Could you not combine it into one method, or are they used separately elsewhere?

frontend/src/APIClients/AuthAPIClient.ts
```diff
@@ -11,7 +11,6 @@ import {
   setLocalStorageObjProperty,
```
DinuWije: Run linters to remove these warning pls

backend/python/requirements.txt
```diff
@@ -68,3 +68,4 @@ uritemplate==3.0.1
 urllib3==1.26.4
 Werkzeug==1.0.1
 zipp==3.4.1
+pytz==2022.7.1
```
DinuWije: This import isn't needed in the requirements.txt file because it's already imported above (in fact, importing it twice causes a build error). Also did you even end up using pytz? I noticed that you imported timezone from both pytz and datetime in the `sign_in_logs_service.py` and I'm not sure which one is being used.

DinuWije (changes requested):
I wasn't able to test the PR because of an unrelated issue that I'm looking into (it's related to our auth roles), but I left some comments. Overall really great work guys, looks super solid! Only a few changes to make and I think this will be good to go. #skillissues

backend/python/app/rest/logs_routes.py
```diff
@@ -0,0 +1,87 @@
...
+    email = request.json["email"]
+    start_date = request.json["start_date"]
+    end_date = request.json["end_date"]
+
+    if email:
+        if type(email) is not str:
+            return jsonify({"error": "email query parameter must be a string"}), 400
+        else:
+            try:
+                user_id = user_service.get_user_by_email(email).id
```
DinuWije, line 39: I think what you are doing is initializing the `user_id` field in the try block and then using it outside of it. This works in python but is generally bad practice (you want to initialize variables in the scope which you are going to use them, which in this case means the variables should be initialized at the top level of the function). TLDR: add something like `user_id = None` at the top level of the function.

DinuWije (approved):

DinuWije:
LGTM demons 🚀😈

PR #36, opened 2023-02-16
Integrate Authy 2FA

### LINKED ISSUE #19: Integrate Duo2FA
## Summary
- Relief staff should only be able to login when they are on-site in the housing building
- The housing building has a single desktop that the relief staff use to make the logs
- The idea to implement the login restriction is to have Duo2FA installed on the desktop, and require that all relief staff enter a code from the app before they can login. 
- Not sure if this is the best approach though, so this task needs some more investigation
- 5 point estimation
## Goals
- [ ] Figure out the best way to implement location based login restrictions (if its not Duo)
- [ ] Figure out if Duo has a native application that can be downloaded and used to generate codes on a computer
- [ ] If Duo is the best option, figure out pricing, API use, and Duo for desktop
- [ ] If Duo is the best option, integrate it with our sign-in flow for Relief Staff
## Implementation
- Start by investigating options to solve the location-based sign-on problem
- For Duo, take a look at the options available and play around with Duo Desktop
- If Duo is suited for our use case, use the Duo API to integrate 2fa auth for the Relief Staff role (check if Duo integrates with Firebase)
## Notes
- This is an investigation heavy task but could be very interesting to develop end-to-end

connor-bechthold:
## Notion ticket link

[Ticket Name](https://github.com/uwblueprint/supportive-housing/issues/19)

## Implementation description
Integrated 2FA using Authy. Basic user flow goes as so:
- User tries to login using normal `/auth/login` route
- Instead of returning the authenticated user, the route will return the following body:
`{requires_two_fa: Boolean, auth_user: User or None}`
- If the user is a relief user, `requires_two_fa` will be set to True and no auth user will be returned, since we don't want an access token unless they're 2FA verified 
- Otherwise, the auth user is returned and the user logs in as normal
- Once the user types in the passcode from the Authy app, they will POST `/auth/twoFa?passcode={passcode}`
- Here, the backend calls the Verify API and verified whether the provided passcode in the query string is valid or not
- If it's valid, it returns the auth_user and logs the user in

Important note: I made the decision to 2FA every time the user has to login. An alternative was to request 2FA even if the user had previously logged in and still had a valid access/refresh token, but I figured that might get extremely annoying. 

Another note: I didn't write any tests for this yet (@Safewaan pls send help), but instead updated the current tests slightly so they don't break with these changes. I figured getting this PR up and reviewed ASAP was more important.

## Steps to test
I want to apologize in advance to whoever has to test this works
 
1. If you don't already have one, [create a new Twilio account](https://www.twilio.com/try-twilio) under your Blueprint email (the free trial suffices, and never ends so you won't be charged)
2. [Create a verify service](https://www.twilio.com/console/verify/services) in the console. You don't need to toggle any of the verification channels, as we're just using TOTP
3. You first need to create a Factor, which is essentially a user that will be accessing this service. The easiest way to do this is through a cURL command:
```
curl -X POST "https://verify.twilio.com/v2/Services/VAXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXX/Entities/ff483d1ff591898a9942916050d2ca3f/Factors" \
--data-urlencode "FriendlyName=Taylor's Account Name" \
--data-urlencode "FactorType=totp" \
-u $TWILIO_ACCOUNT_SID:$TWILIO_AUTH_TOKEN
```
- `VAXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXX` is the ID of the service you made, which you can find here:
Screen Shot 2023-03-12 at 2 10 07 PM

- `ff483d1ff591898a9942916050d2ca3f` is a unique ID that will be associated with your user. This can be a randomly generated from some magical website you find
- The FriendlyName is the name of your user which can be anything
- Don't change FactorType
- `$TWILIO_ACCOUNT_SID` and `$TWILIO_AUTH_TOKEN` are found under `Account > Auth tokens & API Keys`

4. After this hopefully works you will get a response like this:
```
{
  "sid": "YFXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXX",
  "account_sid": "ACXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXX",
  "service_sid": "VAXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXX",
  "entity_sid": "YEXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXX",
  "identity": "ff483d1ff591898a9942916050d2ca3f",
  "binding": {
    "secret": "GEZDGNBVGY3TQOJQGEZDGNBVGY3TQOJQ",
    "uri": "otpauth://totp/test-issuer:John%E2%80%99s%20Account%20Name?secret=GEZDGNBVGY3TQOJQGEZDGNBVGY3TQOJQ&issuer=test-issuer&algorithm=SHA1&digits=6&period=30"
  },
  "date_created": "2015-07-30T20:00:00Z",
  "date_updated": "2015-07-30T20:00:00Z",
  "friendly_name": "Taylor's Account Name",
  "status": "unverified",
  "factor_type": "totp",
  "config": {
    "alg": "sha1",
    "skew": 1,
    "code_length": 6,
    "time_step": 30
  },
  "metadata": null,
  "url": "https://verify.twilio.com/v2/Services/VAXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXX/Entities/ff483d1ff591898a9942916050d2ca3f/Factors/YFXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXX"
}
```
5. Now that the Factor is successfully created, you have to verify that they're able to generate and send the correct code. First, download and open the Authy Desktop App and hit the plus sign in the upper right corner of the app. In the box, enter the `binding.secret` token from the Factor creation, and name the application accordingly. Now, you should be able to see a code being generated.
6. Use the following cURL request to verify the user:
```
curl -X POST "https://verify.twilio.com/v2/Services/VAXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXX/Entities/ff483d1ff591898a9942916050d2ca3f/Factors/YFXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXX" \
--data-urlencode "AuthPayload=293412" \
-u $TWILIO_ACCOUNT_SID:$TWILIO_AUTH_TOKEN
```
- All fields are the same as above, with the exception of `YFXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXX`. This can be found in the create Factor response, under `sid`
- The `AuthPayload` will be the code generated by the Authy app which you'll need to copy over from the desktop app
7. If you get a response that includes `status: approved`, you're done the setup! (once again I apologize this is so long, this only needs to be done once and moving forward all users will be under one general Blueprint account
8. Finally update your `.env` file with the following:
```
# Twilio
TWILIO_ENABLED=True
TWILIO_ACCOUNT_SID=Account SID you used before
TWILIO_AUTH_TOKEN=Auth token you used before
TWILIO_SERVICE_SID=Service SID you used before
TWILIO_ENTITY_ID=Entity ID of the user you created and used before
TWILIO_FACTOR_SID=Factor SID of the factor you made and used before
```
9. Test the 2FA process locally. Ensure that support users are brought through this page, while all other users are not.

If anything here doesn't make sense, you can [refer to the docs](https://www.twilio.com/docs/verify/quickstarts/totp)

* 

## Checklist


frontend/src/components/auth/Credentials.tsx
```diff
@@ -0,0 +1,145 @@
...
+            Sign Up
+          </button>
+        </div>
+      </div>
+    );
+  }
+  return <></>;
+};
+
+export default Credentials;
```
safchow, line 146: the filename for this page is a little misleading, as its the login page, but the login page already exists

frontend/src/components/auth/Authy.tsx
```diff
@@ -0,0 +1,72 @@
...
+    let authUser: AuthenticatedUser | null;
+
+    if (token) {
+      authUser = await authAPIClient.twoFaWithGoogle(passcode, token);
+    } else {
+      authUser = await authAPIClient.twoFa(passcode, email, password);
+    }
+
+    localStorage.setItem(AUTHENTICATED_USER_KEY, JSON.stringify(authUser));
+    setAuthenticatedUser(authUser);
```
safchow: if the user inputs the correct token, they're brought back to the login screen (in my testing). Is it because the following is missing?:

```
return <Redirect to={HOME_PAGE} />;
```

frontend/src/components/auth/Authy.tsx
```diff
@@ -0,0 +1,72 @@
...
+              value={passcode}
+              onChange={(event) => setPasscode(event.target.value)}
+              placeholder="passcode"
+            />
+          </div>
+          <div>
+            <button
+              className="btn btn-primary"
+              type="button"
+              onClick={onAuthyClick}
```
safchow: would it be possible to display a toast message to inform the user the token is incorrect/expired?

safchow:
Some Notes:
- Would it be possible to create a Twillio account the team can use for testing and for production? The process is rather lengthy and ideally, we would create an account for clients to use as the setup process requires making changes to the `.env` file.
- I wasn't able to run e2e tests, but taking a quick look it seems difficult to verify if a user with the Relief Staff role can login with a token as the token is changing every 20 secs. Would it be possible to write a test to check if the user runs into the token page? This might be tricky, as it would require `AUTHY_ENABLED` to be set to True for testing. 
- The setup steps were written very well, and I was able to setup the token auth with no issues. Should the steps be moved to Notion? We might need to create a client doc to set up the app on their devices.

Overall, this looks really good to me and its cool that you got something like this working :rocket: I left a few notes for some fixes and possible additions/changes.

safchow (changes requested):

backend/python/app/rest/auth_routes.py
```diff
@@ -35,13 +37,75 @@
...
+            auth_dto = auth_service.generate_token_for_oauth(request.json["id_token"])
+        else:
+            auth_dto = auth_service.generate_token(
+                request.json["email"], request.json["password"]
+            )
+        response = {"requires_two_fa": False, "auth_user": None}
+
+        if os.getenv("AUTHY_ENABLED") == "True" and auth_dto.role == "Relief Staff":
+            response["requires_two_fa"] = True
+            return jsonify(response), 200
```
DinuWije: What I understand from this code is that the `auth_user` field is defined in the /twoFa route if 2FA is required, so it should be returned null here. Is that correct?

frontend/src/APIClients/AuthAPIClient.ts
```diff
@@ -4,7 +4,7 @@ import {
   OperationVariables,
```
DinuWije, line 4: Run linters pls

frontend/src/APIClients/AuthAPIClient.ts
```diff
@@ -14,28 +14,59 @@ import {
...
+      { email, password },
+      { withCredentials: true },
+    );
+    return data;
+  } catch (error) {
+    return null;
+  }
+};
+
+const twoFaWithGoogle = async (
```
DinuWije, line 60: what's the diff between the `twoFa` method and the `twoFaWithGoogle` method?

frontend/src/components/auth/Credentials.tsx
```diff
@@ -0,0 +1,145 @@
...
+
+  const onLogInClick = async () => {
+    const loginResponse: LoginResponse = await authAPIClient.login(
+      email,
+      password,
+    );
+    if (loginResponse) {
+      const { requiresTwoFa, authUser } = loginResponse;
+      if (requiresTwoFa) {
+        setToggle(!toggle);
```
DinuWije, line 51: what does this toggle do? Does it have to be named `toggle` or can we change it to something more descriptive?

frontend/src/components/auth/Credentials.tsx
DinuWije: Yeah, I'm a little confused by what the toggles do, could u explain pls? At the end of the day if it works, it works (terrible coding practice but this is Blueprint lol), but I wanna make sure the code is readable if other devs pick it up later.

DinuWije (changes requested):
Really great work Connor, this is a sick PR!! A lot of stuff done very well in here, just a few small changes here and there. I know you didn't have time before reading week, but if we could get a test or two in for this that would be great (I wouldn't spend more than one work session on the tests + wtv small changes safwaan and I requested). Also please update the [Notion API Page](https://www.notion.so/uwblueprintexecs/REST-Endpoints-05ce60312bb943439dfda42bb1318536) GREAT JOB

frontend/src/APIClients/AuthAPIClient.ts
connor-bechthold: After the `twoFa` process finishes, the auth token is fetched through Firebase. The `twoFa` authenticates through user and password, while the`twoFaWithGoogle` authenticates with the `idToken` that Google provides. I'm pretty sure we're also doing this with Microsoft so as long as it integrates with Firebase, it shouldn't be too bad to integrate.

frontend/src/components/auth/Credentials.tsx
connor-bechthold: Toggling is explained above

frontend/src/components/auth/Authy.tsx
connor-bechthold: Changed this, could you try again when you get the chance?

frontend/src/components/auth/Authy.tsx
connor-bechthold: Yessir

frontend/src/components/auth/Credentials.tsx
connor-bechthold: The reason for the toggle is so I can cache the username/password/idToken used when logging in. The flow that I designed is like this:
- The `Login` component stores all the state of the login (username, password, etc.) which is passed as props to the two child components, `Credentials` and `Authy`. `toggle` is used to determine which of these components is rendered
- By default, the `Credentials` component renders, which is where you login. 
  - If twoFa isn't required, the auth token will be fetched and the user will be sent to the home page. 
  - If twoFa is required, we don't want to return the auth token and instead trigger the `toggle`, which will render the `Authy` component. Then, once twoFa is completed, the cached credentials in state can easily be sent

I was actually stuck on this for a while...my first thought was to use Router to push to an Authy component, however, there really wasn't a good way to pass the state along. The only real options were adding it to a global context (seems unsafe), or passing it to the new window location state as outlined [here](https://stackoverflow.com/questions/70579317/how-to-pass-state-to-another-component-using-usehistory) (Also seems unsafe to store in window state? Not 100% sure though, may be a viable option). I don't have a ton of frontend knowledge, so if anyone can clarify/verify this that would be great!

backend/python/app/rest/auth_routes.py
connor-bechthold: Yep, that's correct. I defined a default response body above this that sets `auth_user` to `None` by default:
Screen Shot 2023-03-12 at 3 50 36 PM

connor-bechthold:
> Some Notes:
> 
> * Would it be possible to create a Twillio account the team can use for testing and for production? The process is rather lengthy and ideally, we would create an account for clients to use as the setup process requires making changes to the `.env` file.
> * I wasn't able to run e2e tests, but taking a quick look it seems difficult to verify if a user with the Relief Staff role can login with a token as the token is changing every 20 secs. Would it be possible to write a test to check if the user runs into the token page? This might be tricky, as it would require `AUTHY_ENABLED` to be set to True for testing.
> * The setup steps were written very well, and I was able to setup the token auth with no issues. Should the steps be moved to Notion? We might need to create a client doc to set up the app on their devices.
> 
> Overall, this looks really good to me and its cool that you got something like this working 🚀 I left a few notes for some fixes and possible additions/changes.

- For sure, the goal is to have one account that everyone can use under a general BP email so this whole setup only needs to be done once
- That seems like the easiest way, I'll investigate further
- @DinuWije thoughts? This setup should only ever need to be done once, but maybe it's good to have for future reference?

backend/python/app/rest/auth_routes.py
```diff
@@ -35,13 +36,83 @@
...
+
+        if challenge.status != "approved":
+            return (
+                jsonify({"error": "Invalid passcode."}),
+                400,
+            )
+
         auth_dto = None
         if "id_token" in request.json:
             auth_dto = auth_service.generate_token_for_oauth(request.json["id_token"])
```
DinuWije: I don't LOVE that we are doing two separate calls to `generate_token_for_oauth` (once for `/login` and another for this route). We could potentially have two separate login pages (one for Admin, one for Relief Staff) to make sure that only one call is being done every time, however this might be overkill and will probably require some refactoring which might not be worth it in the end. Not saying its something we **have** to do, but I do think it's worth discussing.

backend/python/requirements.txt
```diff
@@ -47,6 +47,7 @@ py==1.10.0
 pyasn1==0.4.8
 pyasn1-modules==0.2.8
 pycparser==2.20
+PyJWT==2.6.0
```
DinuWije: Please merge in the latest main and reupdate the `requirements.txt`

frontend/src/components/auth/Credentials.tsx
DinuWije: Huh that sounds like it makes sense to me, but I also don't have a ton of frontend exp so @juthikahoque could u take a look at this as well? I made a suggestion before to split the sign in flow into two separate flows for admin & relief staff, that might solve the issue you are describing (but again, it could also add unneeded complexity).

frontend/src/components/auth/Authy.tsx
```diff
@@ -0,0 +1,81 @@
...
+  }
+
+  if (toggle) {
+    return (
+      <div style={{ textAlign: "center" }}>
+        <h1>Authy</h1>
+        <form>
+          <div>
+            <p>Enter token from Authy App</p>
+            <input
```
safchow, line 59: hitting enter here seems to break the page

safchow (approved):
lgtm! left a comment on minor bug

frontend/src/components/auth/Authy.tsx
connor-bechthold: fixed good catch!

DinuWije (approved):
KILLER PR!!! LGTM DEMON!!! 🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀

PR #37, opened 2023-02-23
juthika/send users invite link

### LINKED ISSUE #22: User invite link
## Summary
- The client will want to make the platform invite-only
- 5 point estimation
## Goals
- [ ] Admins can call and API endpoint to invite a user to join the platform (an email is sent with a link to join)
- [ ] User invite links expire after 24 hours
- [ ] Admin can specify user role on invite
## Implementation
- Create an API endpoint to send an email to the invited user
- Save the invited user's credentials somewhere (ie. in the user table or consider making an invited user table, but this might be overkill)
- Configure the sign up page so only invited users can sign up

### LINKED ISSUE #67: Invited Users Modifications
**Is your feature request related to a problem? Please describe.**

1. add a column to the users table that marks the user as "deactivated" if we want to delete them (instead of deactivated)
2. add a foreign key relation to the users table to match with the invited_users id
3. if we delete a user through the delete user route then we should add a call to delete from the invited users table

**Describe the solution you'd like**

**Describe alternatives you've considered**

**Additional context**

juthikahoque:
## Notion ticket link

[Ticket Name](https://www.notion.so/uwblueprintexecs/Task-Board-db95cd7b93f245f78ee85e3a8a6a316d)

## Implementation description
* modified users table to add email & user_status columns
* modified users service to create a user entry without an auth_id when a user is invited & to activate a user when they sign up
* added routes for getting/posting
* connected to FE with routes and displayed button to allow you to invite users
* prevents user from logging in if they're not invited
* also made it so that when you login, it will add you to the users table as whatever role the admin invited you as (this can be done at localhost:3000/invite-users)
Screen Shot 2023-05-29 at 11 38 49 PM

## Steps to test
1. pull branch
2. do flask db upgrade
3. add your email into the database as follows: `insert into users (first_name, last_name, role, user_status, email) values ('your first name', 'your last name', 'Admin', 'INVITED', 'your email');`
4. test if your email is invited by signing in again at the /login url (if a "user not found" popup doesn't prevent you from signing in then you're invited!)

* 

## Checklist


backend/python/migrations/versions/0cc810be5494_add_invited_users_db.py
```diff
@@ -0,0 +1,33 @@
...
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
+    # ### commands auto generated by Alembic - please adjust! ###
+    op.create_table('invited_users',
+    sa.Column('id', sa.Integer(), nullable=False),
+    sa.Column('email', sa.String(), nullable=False),
+    sa.Column('role', sa.Enum('Admin', 'Regular Staff', 'Relief Staff', name='roles'), nullable=False),
```
DinuWije: Was this migration working for you locally? It fails for me. I'm pretty sure why it's failing is because the roles enum exists already in our database, so a small change to the migration might be needed.
```suggestion
   sa.Column('role', postgresql.ENUM('Admin', 'Regular Staff', 'Relief Staff', name='roles', create_type=False), nullable=False),
```
With a `from sqlalchemy.dialects import postgresql` at the top

DinuWije: This basically just lets postgres know not to recreate the enum. [source](https://stackoverflow.com/questions/63461381/how-to-use-an-existing-sqlalchemy-enum-in-an-alembic-migration-postgres)

DinuWije (changes requested):
Initial quick review:
Great work on this, looks like a lot of logic went into this feature! I haven't looked at the code yet, but had some issues running your branch so I'll leave an initial PR review and we can resolve any issues tgt! A couple questions:
1. did the migration run fine for u? I had an issue with it (commented the fix in ur pr)
2. I’m able to sign up for an account without an invite link, is this happening for you as well? Can we get someone else to test this?
3. We can resend invite links to people who already are in the system. Can we do a quick check for this (lower priority than the other two)?
4. How did you implement the 1-hour expiry for the links (so I know what to look for in the PR)?
5. Once this PR is merged, does it still make sense to have email verification? Will we need to get rid of this from auth service and Firebase (this might be a lot of work, I'm not sure)?

backend/python/migrations/versions/0cc810be5494_add_invited_users_db.py
juthikahoque: i think it worked for me when i did a flask db downgrade base and then upgrade again, but i can add this in and double check if it works!

juthikahoque: > Initial quick review: Great work on this, looks like a lot of logic went into this feature! I haven't looked at the code yet, but had some issues running your branch so I'll leave an initial PR review and we can resolve any issues tgt! A couple questions:
> 
> 1. did the migration run fine for u? I had an issue with it (commented the fix in ur pr)
> 2. I’m able to sign up for an account without an invite link, is this happening for you as well? Can we get someone else to test this?
> 3. We can resend invite links to people who already are in the system. Can we do a quick check for this (lower priority than the other two)?
> 4. How did you implement the 1-hour expiry for the links (so I know what to look for in the PR)?
> 5. Once this PR is merged, does it still make sense to have email verification? Will we need to get rid of this from auth service and Firebase (this might be a lot of work, I'm not sure)?

- so far i fixed the first three items if you want to take a look! 
- for the one hour expiry, there isn't a way i specifically implemented it (i just took the text from the verification link email which i think is the one that expires in an hour) - i actually wanted to talk to you about that bc i'm not sure if i can implement one hour expiration yet until we figure out what the link should be since right now it's just a link to our localhost:3000 page. on the other hand tho, we do double check that the email the person enters is an invited email, so not sure if that defeats the purpose of the link expiring? bc in that case we could just have it lead to the login page instead of it being like a unique protected page or smthg bc even if someone random tries to log in they won't be able to if they're not invited
- in terms of email verification, i don't think we'll need it if we have sign in verification because it defeats the purpose to verify twice i think in my opinion, but it could be good to just make sure the account was created properly

backend/app/rest/invite_users_routes.py
```diff
@@ -0,0 +1,136 @@
...
+            except Exception as e:
+                error_message = getattr(e, "message", None)
+                return (
+                    jsonify({"error": (error_message if error_message else str(e))}),
+                    500,
+                )
+
+
+@blueprint.route("/", methods=["POST"], strict_slashes=False)
+# @require_authorization_by_role({"Admin"})
```
safchow: we want to restrict this to admin only

backend/app/rest/invite_users_routes.py
```diff
@@ -0,0 +1,136 @@
...
+        invited_user = CreateInvitedUserDTO(**request.json)
+        invited_user_service.create_user(invited_user)
+        email = invited_user_service.send_email_sign_in_link(invited_user.email)
+        return jsonify({"user": invited_user.__dict__, "email": email}), 201
+    except Exception as e:
+        error_message = getattr(e, "message", None)
+        return jsonify({"error": (error_message if error_message else str(e))}), 500
+
+@blueprint.route("/", methods=["DELETE"], strict_slashes=False)
+# @require_authorization_by_role({"Admin"})
```
safchow: we want to restrict this to admin only

safchow (changes requested):
super awesome you got invites working! The only changes are to make some of the APIs admin only. Also if you can, please add the documentation for the table and APIs in the notion.

safchow (approved):
lgtm!

DinuWije (dismissed):
Can we have the user deleted from the invited users table when the user itself is deleted (ie when the users/DELETE endpoint is called)? Want to prevent deleted users from recreating their accounts.

juthikahoque:
> Can we have the user deleted from the invited users table when the user itself is deleted (ie when the users/DELETE endpoint is called)? Want to prevent deleted users from recreating their accounts.

@DinuWije i can add that in but i can't test it bc rn i think if we delete a user it wouldn't return an error because the log records table references users.id as a foreign key (+ some other tables maybe) - should i add in functionality to delete rows in log_records where the users.id is deleted? i'm not sure if that's what we want tho

DinuWije:
> > Can we have the user deleted from the invited users table when the user itself is deleted (ie when the users/DELETE endpoint is called)? Want to prevent deleted users from recreating their accounts.
> 
> @DinuWije i can add that in but i can't test it bc rn i think if we delete a user it wouldn't return an error because the log records table references users.id as a foreign key (+ some other tables maybe) - should i add in functionality to delete rows in log_records where the users.id is deleted? i'm not sure if that's what we want tho

@juthikahoque @Safewaan interesting issue. I think the soln safwaan suggested was: instead of actually deleting users, we set the user to "Inactive" so they are unable to login (unless, say they are reinvited into the platform). This is adding lots of complexity however (new user status, new flow for invite links, new logic for creating an account/checking if one exists). Not sure what you guys think we could do here. I don't really want to delete rows for a deleted user (these log records are probably important). Maybe an alternative approach is linking log records that have their initial creator deleted to an 'Account Deleted' default account? Let me know if that makes sense.

safchow:
> > > Can we have the user deleted from the invited users table when the user itself is deleted (ie when the users/DELETE endpoint is called)? Want to prevent deleted users from recreating their accounts.
> > 
> > 
> > @DinuWije i can add that in but i can't test it bc rn i think if we delete a user it wouldn't return an error because the log records table references users.id as a foreign key (+ some other tables maybe) - should i add in functionality to delete rows in log_records where the users.id is deleted? i'm not sure if that's what we want tho
> 
> @juthikahoque @Safewaan interesting issue. I think the soln safwaan suggested was: instead of actually deleting users, we set the user to "Inactive" so they are unable to login (unless, say they are reinvited into the platform). This is adding lots of complexity however (new user status, new flow for invite links, new logic for creating an account/checking if one exists). Not sure what you guys think we could do here. I don't really want to delete rows for a deleted user (these log records are probably important). Maybe an alternative approach is linking log records that have their initial creator deleted to an 'Account Deleted' default account? Let me know if that makes sense.

like @juthikahoque mentioned, deleting users wouldn't work because they're associated to log records as a foreign key. It would work if we deleted logs with the user, but in the case an employee leaves, we want to retain the logs that they created in the records so its not a feasible solution. We can link it to an default "Account Deleted" user, but this takes away info from the log record (specifically who wrote it) which may be relevant for the log

If we go with the idea of adding statuses for users (active if they're with the org, inactive if they've left) this might remove some headaches with the cases in which an employee leaves. In the employee directory, we can add functionality to allow admins to set the status of staff to either active/inactive. If the employee returns to the organization, they can be set back to active. Active means they can access the platform with their creds, inactive means they cannot. This would also remove the process of having to reinvite employees that return. I think if we flesh it out a little more, it can be functional but it's super manual and it's all assumptions on our end so we need to also double check with SHOW to figure out the solution they would prefer. @juthikahoque @DinuWije what are your thoughts?

DinuWije:
> If we go with the idea of adding statuses for users (active if they're with the org, inactive if they've left) this might remove some headaches with the cases in which an employee leaves. In the employee directory, we can add functionality to allow admins to set the status of staff to either active/inactive. If the employee returns to the organization, they can be set back to active. Active means they can access the platform with their creds, inactive means they cannot. This would also remove the process of having to reinvite employees that return. I think if we flesh it out a little more, it can be functional but it's super manual and it's all assumptions on our end so we need to also double check with SHOW to figure out the solution they would prefer. @juthikahoque @DinuWije what are your thoughts?

The solution you suggested should work, but my main concern with it is complexity to implement (seems like a lot of extra work for an edgecase), but like you said if we want to preserve information about who wrote the log, this MIGHT be our only option (maybe we ask SHOW staff if this is actually important to them or not).

One question that comes up in my mind is why are we strong the foreign key for the user that created it in the first place? Is it only for when we filter the log records? If that's the case, can we just save the employee id (or name, etc.) as a string in the log record instead of a foreign key? just a thought. @juthikahoque @Safewaan

safchow:
> > If we go with the idea of adding statuses for users (active if they're with the org, inactive if they've left) this might remove some headaches with the cases in which an employee leaves. In the employee directory, we can add functionality to allow admins to set the status of staff to either active/inactive. If the employee returns to the organization, they can be set back to active. Active means they can access the platform with their creds, inactive means they cannot. This would also remove the process of having to reinvite employees that return. I think if we flesh it out a little more, it can be functional but it's super manual and it's all assumptions on our end so we need to also double check with SHOW to figure out the solution they would prefer. @juthikahoque @DinuWije what are your thoughts?
> 
> The solution you suggested should work, but my main concern with it is complexity to implement (seems like a lot of extra work for an edgecase), but like you said if we want to preserve information about who wrote the log, this MIGHT be our only option (maybe we ask SHOW staff if this is actually important to them or not).
> 
> One question that comes up in my mind is why are we strong the foreign key for the user that created it in the first place? Is it only for when we filter the log records? If that's the case, can we just save the employee id (or name, etc.) as a string in the log record instead of a foreign key? just a thought. @juthikahoque @Safewaan

You're definitely correct, it is a lot of work because we have to create a new column in users that tracks the status of the user, then also update the login API to only allow users who are active to have permission to log in, and then also write APIs to allow users to switch statuses. Its doable though, and I think it's a good feature to have if SHOW needs it. Especially if they have notable employee turnover. But, this would be oos for this PR

we could take that approach as well, but it might cause some data sync issues (for example, if a user changes their name, the logs they've written won't update, so it might cause some confusion as to who the original author of the log is). Even though there isn't functionality to change your name or any information of a user once they are added to the db, I think we should still maintain the relational approach to our data for good standards.

juthikahoque:
> > > If we go with the idea of adding statuses for users (active if they're with the org, inactive if they've left) this might remove some headaches with the cases in which an employee leaves. In the employee directory, we can add functionality to allow admins to set the status of staff to either active/inactive. If the employee returns to the organization, they can be set back to active. Active means they can access the platform with their creds, inactive means they cannot. This would also remove the process of having to reinvite employees that return. I think if we flesh it out a little more, it can be functional but it's super manual and it's all assumptions on our end so we need to also double check with SHOW to figure out the solution they would prefer. @juthikahoque @DinuWije what are your thoughts?
> > 
> > 
> > The solution you suggested should work, but my main concern with it is complexity to implement (seems like a lot of extra work for an edgecase), but like you said if we want to preserve information about who wrote the log, this MIGHT be our only option (maybe we ask SHOW staff if this is actually important to them or not).
> > One question that comes up in my mind is why are we strong the foreign key for the user that created it in the first place? Is it only for when we filter the log records? If that's the case, can we just save the employee id (or name, etc.) as a string in the log record instead of a foreign key? just a thought. @juthikahoque @Safewaan
> 
> You're definitely correct, it is a lot of work because we have to create a new column in users that tracks the status of the user, then also update the login API to only allow users who are active to have permission to log in, and then also write APIs to allow users to switch statuses. Its doable though, and I think it's a good feature to have if SHOW needs it. Especially if they have notable employee turnover. But, this would be oos for this PR
> 
> we could take that approach as well, but it might cause some data sync issues (for example, if a user changes their name, the logs they've written won't update, so it might cause some confusion as to who the original author of the log is). Even though there isn't functionality to change your name or any information of a user once they are added to the db, I think we should still maintain the relational approach to our data for good standards.

I think that adding another column is probably the best approach because of the foreign key constraint and since most of the tables use the user id as a foreign key already + what safwaan said about their data being modified. i think maybe adding a user id as a foreign key to this table would also be a good idea now that i think of it and adding a column to the users table is probably less work than removing the foreign key relation from all the tables @DinuWije @Safewaan

backend/app/rest/user_routes.py
```diff
@@ -94,19 +95,46 @@ def get_users():
                     500,
                 )
 
+@blueprint.route("/user-status", methods=["GET"], strict_slashes=False)
+@require_authorization_by_role({"Relief Staff", "Regular Staff", "Admin"})
```
safchow: we can't use require_auth_by_role here because this endpoint will just throw a 401 if the user is not logged in and is trying to sign up

safchow: the only issue I see with removing this is that not having it will give information about users that exist on the platform (for example, if an outsider were to input an email of a user on the platform, it would tell us if they have an existing account). one way to avoid this is to create invite links so they can't be shared/accessed from outside

frontend/src/components/pages/LogRecords.tsx
```diff
@@ -60,9 +67,81 @@ const LogRecords = (): React.ReactElement => {
...
+    );
+  };
+  const checkInvitedUser = async () => {
+    const res = await commonApiClient.isUserInvited(invitedEmail);
+    setIsInvited(res);
+  };
   return (
     <div className="page-container">
       <NavigationBar />
+      <Box padding="20px">
```
safchow: since you have a separate page for inviting users, can you remove these changes :D

backend/app/models/user.py
```diff
@@ -10,10 +10,12 @@ class User(db.Model):
     id = db.Column(db.Integer, primary_key=True, nullable=False)
     first_name = db.Column(db.String, nullable=False)
     last_name = db.Column(db.String, nullable=False)
-    auth_id = db.Column(db.String, nullable=False)
+    auth_id = db.Column(db.String, nullable=True)
     role = db.Column(
         db.Enum("Admin", "Regular Staff", "Relief Staff", name="roles"), nullable=False
     )
+    user_status = db.Column(db.Enum("INVITED", "ACTIVE", "DEACTIVATED", name="user_statuses"), nullable=False)
```
safchow: can we switch these to title case? I think you might have to also update the migration file

frontend/src/components/auth/Credentials.tsx
```diff
@@ -41,20 +42,29 @@ const Credentials = ({
...
+            AUTHENTICATED_USER_KEY,
+            JSON.stringify(authUser),
+          );
+          setAuthenticatedUser(authUser);
+        }
       }
+      // Otherwise we can display some sort of error
+    } else {
+      // eslint-disable-next-line no-alert
+      window.alert("user not invited");
```
safchow, line 66: we also need to add a catch for when the user already exists on the app

backend/app/rest/user_routes.py
```diff
@@ -94,19 +95,46 @@ def get_users():
...
+        return jsonify({"user_status": user_status, "email": email}), 201
+    except Exception as e:
+        error_message = getattr(e, "message", None)
+        return jsonify({"error": (error_message if error_message else str(e))}), 500
 
-@blueprint.route("/", methods=["POST"], strict_slashes=False)
-@require_authorization_by_role({"Admin"})
-@validate_request("CreateUserDTO")
+@blueprint.route("/invite-user", methods=["POST"], strict_slashes=False)
+# @require_authorization_by_role({"Admin"})
```
safchow: this shouldn't be commented out

frontend/src/components/pages/InviteUsers.tsx
```diff
@@ -0,0 +1,87 @@
...
+          onChange={(event) => setEmail(event.target.value)}
+          placeholder="username@domain.com"
+        />
+        <Button width="30%" onClick={() => inviteUser()}>
+          Invite user
+        </Button>
+        <Text>
+          Check if user is invited:
+          <Badge colorScheme={isInvited ? "green" : "red"}>
+            {isInvited ? "true" : "false"}
```
safchow: this seems to also be true when the user has already signed up. I don't think this needs to be addressed in this PR, but we can just instead show the user's status

backend/app/models/invited_users.py
```diff
@@ -0,0 +1,34 @@
+from sqlalchemy import inspect
```
safchow: you already mentioned this but leaving this note for me 😅 - invited_users table is no longer needed with the new users table

backend/app/models/user.py
```diff
@@ -10,10 +10,12 @@ class User(db.Model):
     id = db.Column(db.Integer, primary_key=True, nullable=False)
     first_name = db.Column(db.String, nullable=False)
     last_name = db.Column(db.String, nullable=False)
-    auth_id = db.Column(db.String, nullable=False)
+    auth_id = db.Column(db.String, nullable=True)
```
safchow, line 13: we should only allow auth_id to be nullable when the user's status is "Invited". I asked chatgpt (bestie) how to do this, and it looks like this needs to be added:
```
    __table_args__ = (
        CheckConstraint(
            "(user_status = 'INVITED' AND auth_id IS NULL) OR (user_status != 'INVITED' AND auth_id IS NOT NULL)",
            name="check_auth_id_nullable"
        ),
    )
```

safchow: you can probably add it around line 19

backend/app/resources/create_invited_user_dto.py
```diff
@@ -0,0 +1,18 @@
...
+        self.email = kwargs.get("email")
+        self.role = kwargs.get("role")
+        self.first_name = kwargs.get("first_name")
+        self.last_name = kwargs.get("last_name")
+
+    def validate(self):
+        error_list = []
+        if type(self.email) is not str:
+            error_list.append("The email supplied is not a string.")
+        if type(self.role) is not str:
```
safchow: instead of checking if this is a string, should we instead check if its one of the roles?

backend/app/rest/invite_users_routes.py
```diff
@@ -0,0 +1,135 @@
+import os
```
safchow: is this file needed if the invited_users table is being removed?

backend/app/services/implementations/invite_users_service.py
```diff
@@ -0,0 +1,204 @@
+from ..interfaces.invite_users_service import IInviteUserService
```
safchow: does this file get removed as well?

backend/app/services/implementations/user_service.py
```diff
@@ -128,61 +128,113 @@ def get_users(self):
...
-                firebase_user = firebase_admin.auth.create_user(
-                    email=user.email, password=user.password
+            user_entry = User.query.filter_by(email=user.email).first()
+
+            if not user_entry:
+                postgres_invited_user = {
+                    "role": user.role,
+                    "first_name": user.first_name,
+                    "last_name": user.last_name,
+                    "user_status": "INVITED",
```
safchow: this needs to be updated if the table is updated to use title case

backend/app/services/interfaces/invite_users_service.py
```diff
@@ -0,0 +1,75 @@
+from abc import ABC, abstractmethod
```
safchow: does this file go too?

backend/migrations/versions/a2a0a16b6d51_add_status_and_email_to_users_table.py
```diff
@@ -0,0 +1,47 @@
...
+# revision identifiers, used by Alembic.
+revision = 'a2a0a16b6d51'
+down_revision = '82f36cdf325f'
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
+    # ### commands auto generated by Alembic - please adjust! ###
+    user_statuses = postgresql.ENUM('INVITED', 'ACTIVE', 'DEACTIVATED', name='user_statuses')
```
safchow: these would need to be updated to use title case

safchow (changes requested):

backend/app/rest/user_routes.py
safchow: I can make the invite links into a ticket so you don't need to address it with this pr

backend/app/rest/invite_users_routes.py
juthikahoque: nope not needed, all the invited_users stuff will be deleted (service, routes, table, DTO)

backend/app/services/interfaces/invite_users_service.py
juthikahoque: no 😠

juthikahoque: jk yes

backend/app/rest/user_routes.py
juthikahoque: yeah, that makes sense, i think that it should be easy to send an email when the user gets invited

frontend/src/components/auth/Credentials.tsx
juthikahoque: later ticket? i'll add as a todo, also think this would only be on the signup flow not the login

frontend/src/components/pages/InviteUsers.tsx
```diff
@@ -0,0 +1,87 @@
...
+          placeholder="Enter last name"
+        />
+        <Select
+          placeholder="Enter role"
+          value={userRole}
+          onChange={(e) => setUserRole(e.target.value)}
+        >
+          <option value="Relief Staff">Relief Staff</option>
+          <option value="Admin">Admin</option>
+          <option value="Regular Staff">Regular Staff</option>
```
kevin-pierce: I think it's better to throw these into an array and then simply map through the values ; will ensure that there's less copy-pasting if we ever want to extend this list.

```suggestion
         // Somewhere above in this file / in another file
         const ROLE_OPTIONS = ['Relief Staff', 'Admin', 'Regular Staff']

         // Specifically for lines 52-54
         {
               ROLE_OPTIONS.map((role) => <option value={role}>{role}</option>)
         }
```

frontend/src/components/pages/InviteUsers.tsx
```diff
@@ -0,0 +1,87 @@
+import React, { useState } from "react";
+import { Button, Input, Box, Text, Badge, Select } from "@chakra-ui/react";
+
+import NavigationBar from "../common/NavigationBar";
+import commonApiClient from "../../APIClients/CommonAPIClient";
+
```
kevin-pierce, line 7: We can use the following const here and map through the values later when rendering the options (refer to below comment for follow-up code)

```suggestion
const ROLE_OPTIONS = ['Relief Staff', 'Admin', 'Regular Staff']
```

backend/app/models/user.py
```diff
@@ -10,10 +10,19 @@ class User(db.Model):
...
     role = db.Column(
         db.Enum("Admin", "Regular Staff", "Relief Staff", name="roles"), nullable=False
```
kevin-pierce, line 15: Likewise with the change made on the FE, I think it'd be beneficial to export a common enum from somewhere (maybe create a constants file in the `utilities` dir) and use it here, just so we can minimize the number of areas that need to be changed everytime we change the values for the possible roles. Something like this could be implemented (shouldn't be any compilation issues). This simply uses the values in the enum class and maps them as the values of the enum for the column

```suggestion
# Somewhere above in the file / in the /utilities directory
  class RoleOptions(enum.Enum):
      ADMIN = 'Admin'
      RELIEF_STAFF = 'Relief Staff'
      REGULAR_STAFF = 'Regular Staff'


        role = db.Column(
        db.Enum(RoleOptions, values_callable=lambda x: [str(role.value) for role in RoleOptions], name="roles"), nullable=False
    )
```

backend/app/resources/create_invited_user_dto.py
```diff
@@ -0,0 +1,18 @@
...
+        self.email = kwargs.get("email")
+        self.role = kwargs.get("role")
+        self.first_name = kwargs.get("first_name")
+        self.last_name = kwargs.get("last_name")
+
+    def validate(self):
+        error_list = []
+        if type(self.email) is not str:
+            error_list.append("The email supplied is not a string.")
+        if self.role not in ["Admin", "Relief Staff", "Regular Staff"]:
```
kevin-pierce, line 12: If we implement the above enum (and import it here), we can change this to be:

```suggestion
        if self.role not in [role.value for role in RoleOptions]
```

Or if we wanna be extra cringe hyper-runtime-efficient leetcode demons 😈 then we could implement it this way too:
```suggestion
        if self.role not in set(role.value for role in RoleOptions)
```

backend/app/services/implementations/user_service.py
```diff
@@ -128,61 +128,113 @@ def get_users(self):
...
-                "last_name": user.last_name,
-                "auth_id": firebase_user.uid,
-                "role": user.role,
-            }
+    def activate_user(self, user, auth_id=None, signup_method="PASSWORD"):
+        firebase_user = None
+
+        try:
+            print(user.email)
+            print(self.get_user_status_by_email(user.email))
```
kevin-pierce: Do we want to clean these up? I'm assuming they were for debugging

backend/app/services/implementations/user_service.py
```diff
@@ -128,61 +128,113 @@ def get_users(self):
...
+            if(self.get_user_status_by_email(user.email) == "Invited"):
+                if signup_method == "PASSWORD":
+                    firebase_user = firebase_admin.auth.create_user(
+                        email=user.email, password=user.password
+                    )
+                    print("password")
+                    print(firebase_user.uid)
```
kevin-pierce: Ditto on above print cleanup - not sure if these are intentional

frontend/src/components/pages/InviteUsers.tsx
```diff
@@ -0,0 +1,87 @@
...
+        <Input
+          width="70%"
+          type="email"
+          value={invitedEmail}
+          onChange={(event) => setInvitedEmail(event.target.value)}
+          placeholder="username@domain.com"
+        />
+        <Button width="30%" onClick={() => checkInvitedUser()}>
+          Check if invited
+        </Button>
```
kevin-pierce: Missing a bit of context here ; is this something only used for debugging / for the dev env? 

If it's for prod, maybe we'd want to change the text on the button to be a bit clearer (since we're receiving one of 3 states and not necessarily a binary state)

frontend/src/components/pages/InviteUsers.tsx
```diff
@@ -0,0 +1,87 @@
...
+        </Text>
+        <Input
+          width="70%"
+          type="email"
+          value={invitedEmail}
+          onChange={(event) => setInvitedEmail(event.target.value)}
+          placeholder="username@domain.com"
+        />
+        <Button width="30%" onClick={() => checkInvitedUser()}>
+          Check if invited
```
kevin-pierce: Referring to above comment about whether this is for prod:

```suggestion
          Check user status
```

kevin-pierce (changes requested):
Rest of this LGTM barring those minor nits + cleanup items ; functionality wise it's sound, I just think there are some areas where we could minimize code dupe (specifically for the roles) but yeah otherwise this PR is 🔥 

this PR

frontend/src/components/pages/InviteUsers.tsx
safchow: yeah, originally this stuff was on the homepage for testing but it got moved to its own page, so need to clean this up. It'll also be good to refer to when we actually need to implement it (since we can reuse the api calls)

frontend/src/components/pages/LogRecords.tsx
```diff
@@ -7,12 +7,19 @@ import {
   Th,
   Td,
   TableContainer,
+  Button,
```
safchow: if these are unused, you can remove these imports

frontend/src/components/pages/LogRecords.tsx
```diff
@@ -7,12 +7,19 @@ import {
...
+  Text,
+  Badge,
+  Select,
 } from "@chakra-ui/react";
 
 import NavigationBar from "../common/NavigationBar";
 import CreateLog from "../forms/CreateLog";
 import SearchAndFilters from "../common/SearchAndFilters";
 import { LogRecord } from "../common/types/LogRecord";
+import commonApiClient from "../../APIClients/CommonAPIClient";
```
safchow: same here, remove if unused :D

frontend/src/components/pages/InviteUsers.tsx
safchow: we won't actually be using this code for the feature, the front-end will be redone once we have the designs. Though having a list and mapping through it makes more sense than an option component for each item 👍 so we can make note of that later

backend/app/models/user.py
safchow: not sure if this will mesh well with the migration files, since I think those don't need dependencies to run. Adding this change in might break it, but I also don't know what I'm talking about so 😵

safchow (approved):
LGTM (let's get this money) 💸 

Thanks for working as a part time at SHOW and getting this done, I really appreciate it 🥺

backend/app/models/user.py
juthikahoque: hmm @kevin-pierce i think it already has an enum of type roles created in the db, so anytime we use / reference it in the table it doesn't actually create the new enum type of roles

juthikahoque: or at least that's the way it is in the db, and yeah i agree w safwaan that it could cause a lot of conflicts in the migration files so may not be the most worth what do you think?

kevin-pierce: Yeah fair enough - that wasn't something I personally tested when I suggested that change, so I think it's fine to leave as is if it could cause trouble with the migration files

kevin-pierce (approved):
TBH I kinda _don't_ wanna approve this since the PR's 4 month birthday 🎂 🎉 🥳  is in about 10 days, but if we _have_ to... 😞

juthikahoque:
> TBH I kinda _don't_ wanna approve this since the PR's 4 month birthday 🎂 🎉 🥳 is in about 10 days, but if we _have_ to... 😞

LMFAOOOO if u didn't approve it would've acc been on sight 🥊

PR #58, opened 2023-04-05
Create Log Form

### LINKED ISSUE #40: Create Log Component
## Summary
- We need a component which allows staff to create new event logs
- This component should have all fields needed to record logs and make a query to save the log

Image

Blocked by (#28)
## Goals
- [ ] Frontend should match design specifications on Figma
- [ ] Component should make an API call to save the new log to the database
- [ ] User should not be able to submit the component if all the required fields are not filled in

## Implementation
- Consider using a form library (discuss with dev team)
- Confirm exact functionality of the feature with product team

jwangsa:
## Notion ticket link

[Create Log Component](https://github.com/uwblueprint/supportive-housing/issues/40)

## Implementation description
* Created a new component, `CreateLog` that serves as a form to create new logs on the application.
* Used `react-select` library for select and multi-select components
* Used `chakra-dayzed-datepicker` for date picker component
* Set notes column as a mandatory value in the DB
* Styled buttons as per Figma design system (other components need to be designed as well, this will be a task in the future)

### Setup
1. Kill running docker containers
2. Start docker containers. If there are errors (specifically related to missing imports, do a clean build of the deployment)
3. If there are still errors, run `yarn install` in the `frontend` and try again.

### Input Tests
1. Verify the employee select is locked.
2. Verify that the default date and time is the current date and time.
4. Verify that clicking on the date input opens a calendar, where the user can select a date.
5. Verify that selecting another date will update the value in the form.
6. Verify the time input can be changed. If you're on Chrome, I think a fancy dropdown appears. Otherwise, you have to use manual input to set the time.
7. Verify that only one building can be selected.
8. Verify that only one resident can be selected.
9. Verify that multiple tags can be selected.
10. Verify that selected tags can be removed.
11. Verify that an employee can be selected for the "Attention To" input.
12. Verify the notes input functions and updates as the user types.
13. Verify that the checkbox can be selected.

### Error Tests
1. Open a new form.
2. Hit submit, and verify that the required fields are outlined in red, indicating to the user that they must be given.
3. Fill each one out at a time, and verify that the form does not submit until all required fields are given. Additionally, verify that the correct inputs are highlighted if they are missing an input.
4. Verify that an invalid time cannot be given (i.e. erase the time).

### Form Tests
1. Open a new form and fill it out with test data.
2. Hit cancel, and verify that when you open the form again, all the inputs are blank.
3. Verify that when you press submit and there are no errors, the states of the inputs are all correct. This can be done by uncommenting the console.log lines in the `handleSubmit` function in `CreateLog.tsx`. There are two blocks- one for the states, and the error states.
4. After you hit submit, verify that the form appears with all the inputs reset (and the date and time updated if some time has passed).
5. Verify buttons are styled correctly, and change accordingly to the designs [here](https://www.figma.com/file/Gs2ZhbdvW8aLsXDmFO6rky/SHOW-Design-System?type=design&node-id=40%3A2438&t=HEC2YLB7pRtOCdK6-1)

### Extra Tests
1. Try messing around with random inputs to see if you can break anything. This is optional- a lot of the testing above is thorough enough but I might not have covered a case that you're thinking of testing. Also, I was literally fixing errors while writing this so uhhh high chance there's a few errors I missed 😅 

* One of the decisions I had to make was that the date input initially was just one input [here](https://www.figma.com/file/4AStM5IGeHpMw94iU19AkG/Homepage?type=design&node-id=571%3A18362&t=Vl2J8bgQOkSCsB2R-1). I looked for a component library that could do this, but also could be styled relatively easily by the team but it was hard to find one, so I went with breaking it apart. Does it flow well from a user perspective? 
* It was also unclear if the resident and attention to inputs should be an input component or a select component. I went with using a select component because it comes built-in with a search, and it makes implementation easier for us (since the APIs that we'll have to write will all be similar to serve similar information to the maps that store data). Is this something that we should reconsider, or are we okay with it being select components as well?
* We need functionality in the future to create new tags. The nice thing about the library I used is that it can be built into the select component itself, so we don't need to create a whole form for it. We just need to double-check if its intuitive for the client and is relatively easy to build on our end.
* I wasn't able to do all of the styling changes (as mentioned above) since that needs to be its own task, and there are some overlaps with another PR we have open. It makes more sense to get this and other PR merged, then start working on the design system. I already started working on it with the button styling. 

Thanks for making it this far, it's one of the bigger PRs we've had 🐶 

## Checklist


frontend/src/components/common/LogComponent.tsx
```diff
@@ -0,0 +1,76 @@
...
+} from "@chakra-ui/react";
+import React, { useState } from "react";
+
+type LogComponentProps = {
+  employee: string;
+};
+
+// eslint-disable-next-line react/prop-types
+const LogComponent: React.FC<LogComponentProps> = ({ employee }) => {
+  const [building, setBuilding] = useState<string>("");
```
safchow: for the `useState` method, you don't need to state the variable type

frontend/src/components/pages/LogRecords.tsx
```diff
@@ -45,17 +59,252 @@ const mockRecords = [
...
+                    >
+                      <BsHouseFill />
+                    </InputLeftElement>
+                    <Select
+                      placeholder="Building No."
+                      value={building}
+                      onChange={handleBuildingChange}
+                      style={{ textIndent: "10px" }}
+                    >
+                      <option value="Building A">Building A</option>
```
safchow: There are three building options:
144, 362, 402

these should be stored in a list and then map through the list to display each option

frontend/src/components/pages/LogRecords.tsx
```diff
@@ -45,17 +59,252 @@ const mockRecords = [
...
+                        alignItems: "center",
+                        justifyContent: "center",
+                        height: "100%",
+                        paddingLeft: "6px",
+                      }}
+                    >
+                      <BsHouseFill />
+                    </InputLeftElement>
+                    <Select
+                      placeholder="Building No."
```
safchow: we can remove the placeholder for this and default the the main building. Rachel needs to check which building is the main building

frontend/src/components/pages/LogRecords.tsx
```diff
@@ -45,17 +59,252 @@ const mockRecords = [
...
+                    >
+                      <BsHouseFill />
+                    </InputLeftElement>
+                    <Select
+                      placeholder="Enter tags"
+                      value={building}
+                      onChange={handleBuildingChange}
+                      style={{ textIndent: "10px" }}
+                    >
+                      <option value="Tag A">Tag A</option>
```
safchow: we want to also store the tags in a list, later they'll be stored into a table but for now a list should work. We also want to select multiple tags (for example, Tag A and Tag C)

frontend/src/components/pages/LogRecords.tsx
```diff
@@ -45,17 +59,252 @@ const mockRecords = [
...
+                        display: "flex",
+                        alignItems: "center",
+                        justifyContent: "center",
+                        height: "100%",
+                        paddingLeft: "6px",
+                      }}
+                    >
+                      <BsFillPersonFill />
+                    </InputLeftElement>
+                    <Input
```
safchow: For the resident, we want this to be a drop down. The input should act as a search and narrow down the results of the dropdown.

frontend/src/components/pages/LogRecords.tsx
```diff
@@ -45,17 +59,252 @@ const mockRecords = [
...
+                      <option value="Tag B">Tag B</option>
+                      <option value="Tag C">Tag C</option>
+                    </Select>
+                  </InputGroup>
+                </FormControl>
+              </Col>
+              <Col>
+                <FormControl mt={4}>
+                  <FormLabel>Attention To</FormLabel>
+                  <InputGroup>
```
safchow: For the attention to input, we want this to be a drop down. The input should act as a search and narrow down the results of the dropdown. To mock the output, just a create list and we can do the back-end routing as well

frontend/src/components/pages/LogRecords.tsx
```diff
@@ -45,17 +59,252 @@ const mockRecords = [
...
+                  value={value}
+                  onChange={handleInputChange}
+                  placeholder="Enter log notes here..."
+                  size="lg"
+                  style={{ resize: "none" }}
+                />
+              </Col>
+            </Row>
+            <Checkbox defaultChecked style={{ paddingTop: "1rem" }}>
+              Flag this Report
```
safchow: By default, this checkbox should be unchecked.

frontend/src/components/common/LogComponent.tsx
```diff
@@ -0,0 +1,76 @@
...
+    <div className="log-entry">
+      <h1>New Log Entry Details</h1>
+      <form onSubmit={handleSubmit}>
+        <FormControl>
+          <FormLabel>Employee</FormLabel>
+          <Input type="text" value={employee} isDisabled />
+        </FormControl>
+        <FormControl mt={4}>
+          <FormLabel>Date</FormLabel>
+          <Input type="text" value={new Date().toString()} isDisabled />
```
safchow: it looks like the input is also taking a time value (Ex: 7:00pm) but we don't input one. We might want to track the time the user creates the log as well, but I would double-check with Rachel. If so, we need to add a way to input the time. If not, the new format should be "mm/dd/yyyy" instead of "mm/dd/yyyy --:-- --"

safchow: the popup calendar should also be styled according to the figma design:
image

it doesn't have to look the exact same, but the key styling should be there (font, font colour, white background, etc).

frontend/src/components/common/LogComponent.tsx
```diff
@@ -0,0 +1,76 @@
...
+    );
+  };
+
+  return (
+    <div className="log-entry">
+      <h1>New Log Entry Details</h1>
+      <form onSubmit={handleSubmit}>
+        <FormControl>
+          <FormLabel>Employee</FormLabel>
+          <Input type="text" value={employee} isDisabled />
```
safchow: would you be able to mock an employee name for now? Just want to make sure it renders correctly with appropriate styling

frontend/src/components/common/LogComponent.tsx
```diff
@@ -0,0 +1,76 @@
...
+      `Employee: ${employee}, Date: ${new Date()}, Building: ${building}, Tags: ${tags}`,
+    );
+  };
+
+  return (
+    <div className="log-entry">
+      <h1>New Log Entry Details</h1>
+      <form onSubmit={handleSubmit}>
+        <FormControl>
+          <FormLabel>Employee</FormLabel>
```
safchow: For required fields, there should be a red asterisk next to the label like so:
image

I believe you can do this in chakra by passing the `isRequired` prop to the FormControl component:
```
            <FormControl
              id="first-name"
              className="body"
              marginTop="16px"
              isRequired
            >
            ...
            </FormControl>
```

frontend/src/components/pages/LogRecords.tsx
```diff
@@ -45,17 +59,252 @@ const mockRecords = [
...
+                </FormControl>
+              </Col>
+            </Row>
+            <Row>
+              <Col>
+                <FormLabel mt={4}>Notes</FormLabel>
+                <Textarea
+                  value={value}
+                  onChange={handleInputChange}
+                  placeholder="Enter log notes here..."
```
safchow: the font size for this text should be the same as the inputs (16px)

frontend/src/index.css
```diff
@@ -37,4 +37,9 @@ code {
   font-weight: 400;
   color: #FFFFFF !important;
   text-decoration: none
+}
+
+.log-entry h1{
+  font-size: 28px;
+  font-weight: 500;
 }
```
safchow, line 84: some styling is missing, like the font colour, which should be #536C6E. The same applies to the icons. There is also an on-focus effect for the text inputs to change the border to #285E61.

safchow (dismissed):
Good job on getting the front end for adding logs done so quickly! I left a few comments on some behaviour for the inputs, as well as some minor changes that should be made.

safchow:
oh also, I would recommend filling out the PR description body with the implementation description and steps to test, makes reviewing easier 😄

frontend/src/App.tsx
safchow, line 1: This change needs to be undone before merging

connor-bechthold:
Couple of things I wanted to point out in the tests I ran with the form (ranked from most important to least important):
- If I submit a form at 4:30 PM, wait for a few minutes, and open the form back up, the starting time is still 4:30PM
- I found the date picker to be a little confusing; days on previous and future months are able to be selected on the current month. It works fine, but is there a way that those days could be greyed out to reduce confusion?
Screen Shot 2023-05-13 at 4 14 16 PM.
- Looking at Figma, should the Submit button be greyed out when all required forms aren't there? Or is the idea to rely on the error message prompts instead

Also other points:
- I like the split date/time better than it all being together
- Select components are good, especially because if our building/employee list grows (unlikely), it's convenient to able to search AND scroll

frontend/src/components/pages/LogRecords.tsx
```diff
@@ -45,17 +47,21 @@ const mockRecords = [
...
     Resident: "MB404",
-    Note: "During security check, MB404 was making some noise. TSW warned her to be quiet. She yelled on TSW behind the door, yelled, and swore (f..uck of......). TSW told her that I will call the police if she continues. Then she came down for laundry.",
+    Note:
+      "During security check, MB404 was making some noise. TSW warned her to be quiet. She yelled on TSW behind the door, yelled, and swore (f..uck of......). TSW told her that I will call the police if she continues. Then she came down for laundry.",
     Employee: "Huseyin",
     Attn_To: "John Doe",
   },
 ];
 
-const LogRecords = (): React.ReactElement => {
```
connor-bechthold: Any reason for removing this?

frontend/src/components/forms/CreateLog.tsx
```diff
@@ -0,0 +1,394 @@
...
+                                    </FormControl>
+                                </Col>
+                            </Row>
+
+                            <Row>
+                                <Col>
+                                    <FormControl mt={4}>
+                                        <FormLabel>Tags</FormLabel>
+                                        <Select
+                                            options={TAGS}
```
connor-bechthold, line 334: If you have too much time on your hands, could put this in state and set it with the test data in a useEffect when the component renders

frontend/src/components/forms/CreateLog.tsx
```diff
@@ -0,0 +1,394 @@
...
+    ModalHeader,
+    ModalOverlay,
+    Text,
+    Textarea,
+} from "@chakra-ui/react";
+
+import { AddIcon } from '@chakra-ui/icons'
+import { SingleDatepicker } from "chakra-dayzed-datepicker";
+import { Col, Row } from "react-bootstrap";
+
```
connor-bechthold, line 29: If you have too much time on your hands, putting these in state and setting them in a useEffect when the component renders would replicate an actual connection to the database more.

frontend/src/components/forms/CreateLog.tsx
```diff
@@ -0,0 +1,394 @@
...
+    const handleTimeChange = (e: React.ChangeEvent<HTMLInputElement>) => {
+        setTime(e.target.value);
+        const timeRegex = /^(0[0-9]|1[0-9]|2[0-3]):[0-5][0-9]$/; // Regex to match time format HH:mm
+
+        // Check to see if the input is valid, prevents an application crash
+        if (timeRegex.test(e.target.value)) {
+            const [hour, minute] = e.target.value.split(":");
+            const updatedDate = new Date(date); // update the time values of the current date state
+            updatedDate.setHours(parseInt(hour, 10));
```
connor-bechthold, line 104: Curious as to why we're also changing hours/minutes here? Shouldn't the time always correspond to the date picked, changing it shouldn't affect the date no?

connor-bechthold (changes requested):
Couple comments to address, looks really good!

frontend/src/components/pages/LogRecords.tsx
safchow: not really, I'll add it back since its good practice to indicate this is a React element

frontend/src/components/forms/CreateLog.tsx
safchow: I have no time 😭 good to note for the future though :D

safchow:
>     * Looking at Figma, should the Submit button be greyed out when all required forms aren't there? Or is the idea to rely on the error message prompts instead

just noticed this- I think it's better from an information architecture perspective to highlight the fields that are required when a submission is made with missing information. This way, the user isn't confused as to whats wrong with their form

frontend/src/components/forms/CreateLog.tsx
safchow: Whenever the date state is updated, the dd/mm/yyyy values change but never the hh:mm value. This updates both the time state, so that it renders correctly, and the date state, so that when it is submitted, it'll be at the correct time. In our db, we store the date value in one column (instead of date and time).

safchow:
>     * I found the date picker to be a little confusing; days on previous and future months are able to be selected on the current month. It works fine, but is there a way that those days could be greyed out to reduce confusion?

I looked into this and it's possible, it just requires the calendar to be built a little more from scratch (the Dayzed library provides it, the component I used is just a template with limited customization). I think its fine to keep it as is for now, and see if its an issue with the clients and address it accordingly.

connor-bechthold (approved):
LGTM, just wanted to point this out again:
`After you hit submit, verify that the form appears with all the inputs reset (and the date and time updated if some time has passed).`

As of now what happens is:
- Open the form at 12:30
- Submit at 12:35
- When I open the form again, the time is still at 12:30

safchow:
> LGTM, just wanted to point this out again: `After you hit submit, verify that the form appears with all the inputs reset (and the date and time updated if some time has passed).`
> 
> As of now what happens is:
> 
>     * Open the form at 12:30
> 
>     * Submit at 12:35
> 
>     * When I open the form again, the time is still at 12:30

weird... it should update whenever the form is opened. Can you try pulling again and lmk if you run into the same issue?

if you do, let me know what browser you're on. It might be related- I'm on Firefox

backend/app/models/log_records.py
```diff
@@ -13,7 +14,7 @@ class LogRecords(db.Model):
     flagged = db.Column(db.Boolean, nullable=False)
     attn_to = db.Column(db.Integer, db.ForeignKey("users.id"), nullable=True)
     # TODO: replace open String fields with VarChar(NUM_CHARS)
-    note = db.Column(db.String, nullable=True)
+    note = db.Column(db.String, nullable=False)
```
phamkelly17, line 17: some backend changes need to be made with this, because there are some clauses that check if a note is there. this doesn't have to be done this PR but just something to note

phamkelly17 (approved):
LGTM

PR #50, opened 2023-03-12
Create residence table in database

helenellyx:
## Notion ticket link

[Database Changes for Residents](https://github.com/uwblueprint/supportive-housing/issues/43)

## Implementation description
* Added residence table in database with the fields 
1. id
2. date_joined
3. status (this has enum type and the specific strings will be finalized later by the design team)

## Steps to test
- if the table has been migrated correctly

## What should reviewers focus on?
* Should I make status have an enum type or just have it as a string type?

## Checklist
- [x] My PR name is descriptive and in imperative tense
- [x] My commit messages are descriptive and in imperative tense. My commits are atomic and trivial commits are squashed or fixup'd into non-trivial commits
- [ ] I have run the appropriate linter(s)
- [x] I have requested a review from the PL, as well as other devs who have background knowledge on this PR or who will be building on top of this PR
- [x] If I have made API changes, I have updated the [REST API Docs](https://www.notion.so/uwblueprintexecs/REST-Endpoints-05ce60312bb943439dfda42bb1318536)
- [x] IF I have made changes to the db/models, I have updated the [Data Models Page](https://www.notion.so/uwblueprintexecs/Data-Models-760f8aa06b244eb0842c079ad77987b0)
- [x] I have updated other Docs as needed

PR #74, opened 2023-05-25
Log Records Seeding Script

### LINKED ISSUE #71: Investigate how to pre-populate DB with fake data
## Summary
Once the homepage is mostly built, it would be helpful to have the db filled with some fake users, as well as some log records for testing. 

Wait till all db changes (#50, #61, #63) are merged.

## Goals
- [ ] Investigate how to fill the DB with fake data (probably needs a separate migration file that can be included depending on an env variable)

connor-bechthold:
## Implementation description
Created a bash script that can be ran to seed the log records table with an arbitrary amount of semi-random records

Add an env file in the seeding directory with these values:
```
#Log Records Table
LOG_RECORD_ROWS=50
EMPLOYEE_ID=1
ATTN_TO_ID=1
```

## Steps to test
1. Run `bash seed.sh` (if you're on Windows. include a -w flag)
2. Check your table to ensure entries are there

Whipped this up with little thought so there's probably room for improvement, but it at least serves as a solid starting point that can be built upon

## Checklist


db-init/seed.sql
```diff
@@ -0,0 +1,24 @@
+-- Seed log records table
+INSERT INTO log_records (
+  employee_id, 
+  resident_first_name, 
+  resident_last_name,
```
safchow: this needs to be updated to resident_id but we need to wait for #50 to be merged first

db-init/seed.sql
```diff
@@ -0,0 +1,24 @@
+-- Seed log records table
+INSERT INTO log_records (
+  employee_id, 
+  resident_first_name, 
+  resident_last_name, 
+  datetime, 
+  flagged, 
+  attn_to, 
+  note, 
+  tags,
```
safchow: same here as well, need to wait for your tags PR to be merged

db-init/seed.sql
```diff
@@ -0,0 +1,24 @@
...
+  resident_first_name, 
+  resident_last_name, 
+  datetime, 
+  flagged, 
+  attn_to, 
+  note, 
+  tags, 
+  building ) 
+
+SELECT
```
safchow: would we able to do something similar to set up residents and a few dummy accounts in the db? It would definitely help set up a testing env fairly quickly and we can include an admin account here and in Firebase (so we don't have to keep creating accounts on fresh runs of the containers)

db-init/seed.sql
```diff
@@ -0,0 +1,24 @@
...
+  1,
+  (ARRAY['John','Jerry','James','Jack'])[floor(random() * 4 + 1)], 
+  (ARRAY['Doe','Show','Flow','Mow'])[floor(random() * 4 + 1)], 
+  NOW() + (random() * (NOW()+'90 days' - NOW())) + '30 days', 
+  (ARRAY[true,false])[floor(random() * 2 + 1)], 
+  1, 
+  (ARRAY['Amazing note','This is a note','Bad note','Decent Note','Cool note'])[floor(random() * 5 + 1)], 
+  ARRAY['tag1', 'tag2'], 
+  (ARRAY['Building A','Building B','Building C'])[floor(random() * 3 + 1)]
```
safchow: buildings are:
144, 362, 402

db-init/seed.sql
connor-bechthold: it can never be escaped

db-init/seed.sql
connor-bechthold: Yes

connor-bechthold:
I think it would be beneficial to at least share these files in Slack so they can be used as a template for other tables

Just so we're not waiting on a merge

db-init/seed.sql
safchow: we can ignore this and merge what you have for now

db-init/seed.sql
```diff
@@ -0,0 +1,24 @@
...
+
+FROM generate_series(1, 50); -- Change this to get a different amount of rows
```
safchow: can we move this to a env file in db-init? I'd like to avoid bi-accidently pushing changes to this file

connor-bechthold: The number of rows?

safchow: yeah

seeding/seed.sh
```diff
@@ -0,0 +1,7 @@
+#!/bin/bash
+
+#Get absolute path to seeding dir
+SCRIPT_DIR=$( cd -- "$( dirname -- "${BASH_SOURCE[0]}" )" &> /dev/null && pwd )
+cat $SCRIPT_DIR/seed.sql | docker exec -i SHOW-database /bin/bash -c 'psql -U postgres -d show_db'
```
safchow: below was an error I ran into running the script on Windows:
```
OCI runtime exec failed: exec failed: unable to start container process: exec: "C:/Program Files/Git/usr/bin/bash": stat C:/Program Files/Git/usr/bin/bash: no such file or directory: unknown
```

safchow: The fix is to double the slashes so that its `//bin//bash`, its something to do with Windows interpreting it as a local path instead of a dir on the Docker container (Windows diff)

Idk if this would effect Linux environments, but if it does we can just have to versions of this file, one for Windows and and one for Linux

safchow (changes requested):

seeding/seed.sh
kevin-pierce: Tested this locally on Mac (UNIX); had no issues running the script with `bash` OR `zsh`, so I think we could proceed with the two files soln. I'm thinking we just add a subdir `/windows` and `/unix` to the `seeding` directory OR prepend the OS to the file names themselves

seeding/seed.sh
```diff
@@ -0,0 +1,7 @@
...
+
+echo "Database seeding complete"
```
safchow, line 53: can you display a error message if the above process fails?

connor-bechthold: you'll know if it fails from the sql logs 

also the toast msg won't ever appear

seeding/seed.sh
```diff
@@ -0,0 +1,53 @@
+#!/bin/bash
+
+#If you're on Windows, run bash seed.sh -w
+#Otherwise, run bash seed.sh
+
+#Import root env file variables
+source ../.env
```
safchow: would it be possible to have an .env file in the seeding dir instead?

safchow (approved):
another W Connor moment 😎 small nit for the .env but I'm good to merge this in

Next session you should go over this PR, how to use it, and how to add to it (for example, creating fake residents), it would be extremely helpful for the team :D

kevin-pierce (approved):
LGTM 🔥

PR #75, opened 2023-05-26
Download Log Records as CSV file on Homepage 

### LINKED ISSUE #52: Export Daily Log to CSV
## Summary
On the log page, there should be a button to allow users to export logs from a given range to a .CSV or PDF file. 

## Goals
- [ ] With a query, pull all logs within the given date
- [ ] Pass the response into a JSON to CSV library
- [ ] The CSV should be downloadable by the user

## Implementation
- Need to investigate how to pull data from JSON to a CSV file

owen-sellner:
## Notion ticket link

[Export Daily Log to CSV](https://github.com/uwblueprint/supportive-housing/issues/52)

## Implementation description
*  Added `return_all` parameter to the Get Log Records API to ignore the pagination when getting records.
* Created a new component, `PrintCSVButton` that prompts the user for a date range which is used to filter the log records.
* Created the function `CSVConverter` to convert JSON to CSV and download the result as a CSV file.

### Setup
1. Set `RESULTS_PER_PAGE` to 1 in `.env` 
2. Create 3 new entries in the `log_records` table with different dates

### API Tests
1. Verify that `localhost:8080/log_records/?return_all=true` returns all 3 log records
2. Verify that `localhost:8080/log_records/?return_all=false` returns the first log record only
3. Verify that `localhost:8080/log_records/?page_number=1` returns the first log record only
4. Verify that `localhost:8080/log_records/` returns the first log record only

### Functional Testing
1. Verify on the homepage that a modal opens when the printer icon is clicked.
2. Verify that the cancel button closes the modal 
3. Verify that the print button downloads a CSV file of all records if a date range is not selected
4. Verify that a date range can be selected 
5. Verify that the clear button removes any selected date range
6. Verify that when a date range is selected the print button downloads a CSV file of records within the date range
7. Verify that when a date range is selected that contains no records the user is shown an alert when the print button is pressed

* Should the user receive a "success" alert when the CSV file is successfully created?
* Should an empty CSV file be downloaded if a date range has no results instead of giving an alert?

## Checklist


backend/app/services/implementations/log_records_service.py
```diff
@@ -98,11 +98,13 @@ def filter_by_flagged(self, flagged):
         print(flagged)
         return f"\nflagged={bool(flagged)}"
 
-    def get_log_records(self, page_number, filters=None):
+    def get_log_records(self, page_number, return_all, filters=None):
         try:
             results_per_page = int(os.getenv("RESULTS_PER_PAGE"))
```
phamkelly17, line 107: can't we move this into the if claise on line 105 as well?

phamkelly17 (approved):
LGTM

backend/app/rest/log_records_routes.py
```diff
@@ -34,22 +34,26 @@ def get_log_records():
     Get RESULTS_PER_PAGE log records. Will return the log records of corresponding to the page you're on. Can optionally add filters.
     """
     page_number = 1
+    return_all = False
     try:
         page_number = int(request.args.get("page_number"))
     except:
```
connor-bechthold, line 40: Want to verify here the point of the nested try/except is because if we're given a page number, we don't care about return all at all? Basically `page_number` takes precedence over `return_all`

backend/app/services/implementations/log_records_service.py
```diff
@@ -153,9 +155,15 @@ def get_log_records(self, page_number, filters=None):
...
+
+            # overwrite json_list to be the subset of data to return if return_all is false
+            if not return_all:
+                json_list = json_list[start_index:end_index]
+
             return {
-                "log_records": json_list[start_index:end_index],
+                "log_records": json_list,
                 "num_results": num_results,
+                "return_all": return_all,
```
connor-bechthold: Is there a point of returning `return_all`

i.e. are we ever going to use it anywhere

frontend/src/APIClients/CommonAPIClient.ts
```diff
@@ -3,31 +3,34 @@ import { getLocalStorageObjProperty } from "../utils/LocalStorageUtils";
...
-      params: {
-        filters: {
-          building,
-          employeeId,
-          attnTo,
-          dateRange,
-          tags,
-          flagged,
+    const { data } = await baseAPIClient.get(
+      `/log_records?return_all=${return_all}`,
```
connor-bechthold: `return_all` can be put as part of `params` 

`params: {filters: {...}, return_all}`

frontend/src/components/common/PrintCSVButton.tsx
```diff
@@ -0,0 +1,146 @@
...
+
+    const data = await commonAPIClient.filterLogRecords(
+      undefined,
+      undefined,
+      undefined,
+      dateRange,
+      undefined,
+      undefined,
+      true, // return all data
+    );
```
connor-bechthold: This looks a bit strange. Doesn't have to be apart of this PR, but what would you think of instead passing an object to this function where the key-value pairs are only the filters we want to include? Then, in the function, any keys that are left out can be replaced by their default. We can easily define a type for this as well

frontend/src/components/common/PrintCSVButton.tsx
```diff
@@ -0,0 +1,146 @@
...
+  };
+
+  const handleClose = () => {
+    setOpen(false);
+  };
+
+  const handleSubmit = async () => {
+    // Use YYYY-MM-DD format
+    const dateRange = selectedDates.map((date) =>
+      date.toISOString().substring(0, 10),
```
connor-bechthold: One thing to call out, `toISOString()` converts the date to UTC (in our case, 5 hours ahead). Just want to ensure this isn't going to cause any weird date issues (only reason I know this is because I spent like a month at my past co-op trying to debug something like this lol)

frontend/src/components/common/PrintCSVButton.tsx
```diff
@@ -0,0 +1,146 @@
...
+    setShowAlert(CSVConverter(data.logRecords) === "error");
+  };
+
+  useEffect(() => {
+    if (showAlert) {
+      setTimeout(() => {
+        setShowAlert(false);
+      }, 3000);
+    }
+  }, [showAlert]);
```
connor-bechthold, line 63: This is really nice, would definitely be open to using this in other areas where we have modal alerts

frontend/src/components/common/SearchAndFilters.tsx
```diff
@@ -65,6 +64,7 @@ const SearchAndFilters = ({ setLogRecords }: Props): React.ReactElement => {
         dateRange,
         tags ? [tags] : [],
         flagged,
+        false, // do not return all data
```
connor-bechthold: Relates back to my comment above, there would be no need to do this if we passed in an object

frontend/src/helper/CSVConverter.tsx
```diff
@@ -0,0 +1,37 @@
+import { LogRecord } from "../components/common/types/LogRecord";
+
+const CSVConverter = (data: any): string => {
```
connor-bechthold: Does the LogRecord[] type exist yet? If so, we should be using it here and not any

frontend/src/helper/CSVConverter.tsx
```diff
@@ -0,0 +1,37 @@
...
+    link.click();
+
+    // Cleanup created object
+    document.body.removeChild(link);
+    URL.revokeObjectURL(url);
+
+    return "success";
+  } catch {
+    return "error";
+  }
```
connor-bechthold, line 34: Nitpick, but personally feel it's easier to just return True or False here

connor-bechthold (dismissed):
Really sick feature! Lot less code needed than I thought which is nice. Also remember to run the linter (if you can't, let me know and I can)

Left some comments. Also, found a (possible?) edge case when selecting dates that I don't think works as expected
Screenshot 2023-05-31 at 10 26 57 PM
Screenshot 2023-05-31 at 10 27 12 PM
Screenshot 2023-05-31 at 10 27 30 PM

Here the log record with date of `March 15th, 2023` wasn't included in the CSV, when I believe it should've. May have to do with the fact that the records I stored had times on them.

frontend/yarn.lock
```diff
@@ -11209,8 +11209,8 @@ typedarray@^0.0.6:
   resolved "https://registry.npmjs.org/typedarray/-/typedarray-0.0.6.tgz"
   integrity "sha1-hnrHTjhkGHsdPUfZlqeOxciDB3c= sha512-/aCDEGatGvZ2BIk+HmLf4ifCJFwvKFNb9/JeZPMulfgFracn9QFcAf5GO8B/mweUjSoblS5In0cWhqpfs/5PQA=="
 
-typescript@^4.9.5:
-  version "4.9.5"
+typescript@^4.1.2:
+  version "4.1.2"
```
kevin-pierce, line 11213: Don't think was intentional ; could we revert this?

frontend/src/components/common/PrintCSVButton.tsx
```diff
@@ -0,0 +1,146 @@
...
+
+const PrintCSVButton = (): React.ReactElement => {
+  const [selectedDates, setSelectedDates] = useState<Date[]>([]);
+  const [showAlert, setShowAlert] = useState(false);
+
+  const handleClear = () => {
+    setSelectedDates([]);
+  };
+
+  const [isOpen, setOpen] = useState(false);
```
kevin-pierce: nit: Can we move all uses of useState to the top of the function, and all handlers below? (i.e. in this case just swap the ordering of these 2)

frontend/src/components/common/PrintCSVButton.tsx
```diff
@@ -0,0 +1,146 @@
...
+          <ModalOverlay />
+          <ModalContent>
+            <ModalHeader>Print CSV File</ModalHeader>
+            <ModalBody>
+              <FormControl>
+                <Flex gap="8px">
+                  <RangeDatepicker
+                    selectedDates={selectedDates}
+                    onDateChange={setSelectedDates}
+                  />
```
kevin-pierce: One thing I think that might be more beneficial in our case is simply copy-pasting this from https://github.com/aboveyunhai/chakra-dayzed-datepicker/tree/main/src/components and leveraging that instead for increased customizability going forward (owner of the pkg also recommends this)

frontend/src/components/common/PrintCSVButton.tsx
```diff
@@ -0,0 +1,146 @@
...
+      <Box>
+        <Modal isOpen={isOpen} onClose={handleClose} size="xl">
+          <ModalOverlay />
+          <ModalContent>
+            <ModalHeader>Print CSV File</ModalHeader>
+            <ModalBody>
+              <FormControl>
+                <Flex gap="8px">
+                  <RangeDatepicker
+                    selectedDates={selectedDates}
```
kevin-pierce, line 84: I think it's a little hard to discern that this is a date-input (maybe we wanna update the [designs too?](https://www.figma.com/file/4AStM5IGeHpMw94iU19AkG/Homepage?type=design&node-id=571-18362&t=CxnlYZB8mx2GbBPe-0)) 

We can add this prop to add some placeholder text so it's more indicative of what this component is for:
Screen Shot 2023-06-04 at 12 46 04 AM




```suggestion
                    selectedDates={selectedDates}
                    propsConfigs={{
                      inputProps: {
                        placeholder: 'DD/MM/YYYY -> DD/MM/YYYY'
                      },
                    }}
```

frontend/src/components/common/PrintCSVButton.tsx
```diff
@@ -0,0 +1,146 @@
...
+      undefined,
+      undefined,
+      undefined,
+      dateRange,
+      undefined,
+      undefined,
+      true, // return all data
+    );
+
+    setShowAlert(CSVConverter(data.logRecords) === "error");
```
kevin-pierce: Could we instead have the CSVConverter fn return `true / false`? Makes this a bit cleaner since we aren't really doing anything with the `error` or `success` otherwise

frontend/src/components/common/PrintCSVButton.tsx
```diff
@@ -0,0 +1,146 @@
...
+
+      <Box>
+        <Modal isOpen={isOpen} onClose={handleClose} size="xl">
+          <ModalOverlay />
+          <ModalContent>
+            <ModalHeader>Print CSV File</ModalHeader>
+            <ModalBody>
+              <FormControl>
+                <Flex gap="8px">
+                  <RangeDatepicker
```
kevin-pierce, line 83: Also not sure how big of a concern this is, but do we want an aria-label on this?

frontend/src/components/common/PrintCSVButton.tsx
kevin-pierce: As an added safety check, just so we don't get any unwanted `Cannot read properties of null` errors

```suggestion
    setShowAlert(!data || CSVConverter(data.logRecords) === "error");
```

kevin-pierce (changes requested):
Overall looks good! Left a few comments, lmk if you have any questions about them

frontend/yarn.lock
safchow: we use version 4.1.2 for ts, I accidently upgraded it to 4.9.5 in an earlier PR but never updated the yarn.lock file and caught it here

kevin-pierce: Ahhh gotcha ok, resolving in that case

frontend/src/components/common/PrintCSVButton.tsx
safchow: there are no designs for this specific modal 😅 I'm good with the suggestion above though

frontend/src/components/common/PrintCSVButton.tsx
safchow: I think its okay to leave it the way it is for now, there is a new ticket to style the date picker (#86) so I can include the link above in that ticket

backend/app/rest/log_records_routes.py
safchow: yeah if there's a `page_number`, `return_all` will not be considered. If there's no `return_all`, the page_number will default to 1

backend/app/services/implementations/log_records_service.py
safchow: not really, think we left this in accidentally when we were testing

backend/app/services/implementations/log_records_service.py
owen-sellner: I am not sure if I understand, which part do you think should be moved?

frontend/src/components/common/PrintCSVButton.tsx
owen-sellner: I think it is just for accessibility. Is there something you would want instead?

frontend/src/components/common/PrintCSVButton.tsx
owen-sellner: This was a good catch, thanks!

frontend/src/components/common/PrintCSVButton.tsx
owen-sellner: I can make sure the date is always in the same time zone as Toronto by doing this instead:
```
date.toLocaleString("fr-CA", {timeZone: "America/Toronto"}).substring(0, 10)
```

Would this fix the issue you are mentioning?

**Note:** "fr-CA" sets the format to yyyy-MM-dd

frontend/src/components/common/PrintCSVButton.tsx
kevin-pierce: Yeah for accessibility ; not sure if it's a large concern, but I think generally we should tack something on for any interactive elements on the page. @Safewaan are there any guidelines for this?

I'm fine with leaving it as is though

kevin-pierce (approved):
Rest of it LGTM! The aria-label thing is very minor, shouldn't block this IMO

frontend/src/components/common/PrintCSVButton.tsx
connor-bechthold: Should be good! This is only being used in Waterloo so it's safe to do that

frontend/src/components/common/types/Filters.ts
```diff
@@ -0,0 +1,9 @@
+export type Filters = {
```
safchow: maybe rename this to LogRecordFilters, since we'll also have filters for employees and residents in the future

owen-sellner: Yeah that makes sense, I will change it. 👍

Issue #72, opened 2023-05-23
Pagination Component on Homepage


safchow:
## Summary
The `GET` request for log_records uses pagination to return a select number of logs. In the .env file, the following line can be updated to change the number:
```
RESULTS_PER_PAGE=50 
```

Ideally, this number should be selected by the user and shouldn't be set in the .env file. The current options are: 10, 25, 50.

Furthermore, the user should also be able to navigate to the next page so that the next set of logs display on the page. On the back-end, passing `page_number` as a query param of the page number will show the next x logs. 

## Goals
- [x] Create front-end components for pagination
- [x] Update log_records to use the number of results the user selects 
- [x] Link back-end and front-end (clicking on the next page shows the next set of results, changing the number of results updates corectly, etc.)

PR #76, opened 2023-05-26
Design system

phamkelly17:
## GH ticket link

[Implement Design System](https://github.com/uwblueprint/supportive-housing/issues/64)

## Implementation description
* Added theming to replace `frontend/src/index.css`. So far the design system covers the navbar and the log records tabl

## Checklist
- [x] My PR name is descriptive and in imperative tense
- [x] My commit messages are descriptive and in imperative tense. My commits are atomic and trivial commits are squashed or fixup'd into non-trivial commits
- [x] I have run the appropriate linter(s)
- [x] I have requested a review from the PL, as well as other devs who have background knowledge on this PR or who will be building on top of this PR
- [x] If I have made API changes, I have updated the [REST API Docs](https://www.notion.so/uwblueprintexecs/REST-Endpoints-05ce60312bb943439dfda42bb1318536)
- [x] IF I have made changes to the db/models, I have updated the [Data Models Page](https://www.notion.so/uwblueprintexecs/Data-Models-760f8aa06b244eb0842c079ad77987b0)
- [x] I have updated other Docs as needed

PR #77, opened 2023-05-31
Update homepage with log records from DB + pagination

### LINKED ISSUE #70: Update Homepage table to render data from DB
## Summary
The log records table on the homepage currently uses mock data. This should be replaced with the log records from the database instead. Some of the backend work is already done in #65. 

## Goals
- [ ] After #65 is merged in, use the `logRecords` state in `LogRecords.tsx` to render the data from the db on the page.
- [ ] There is some investigation involved, as `logRecords.map()` doesn't work, so simply changing `mockRecords.map` isn't the solution.
- [ ] Another note: Date and Time are two different columns on the frontend but they're stored as one column in the db, they should be correctly split into their respective columns when rendered.

kevin-pierce:
## GH ticket link

[Pagination Component on Homepage](https://github.com/uwblueprint/supportive-housing/issues/72)

Related ticket
[Update Homepage table to render data from DB](https://github.com/uwblueprint/supportive-housing/issues/70)

## Implementation description
* Updated `get_log_records` service + associated endpoint to take in `results_per_page` as parameter, allowing for chunked request sizes
* Created `Pagination` component as per functional description + Figma designs
* Extracted code for log records table into `common/LogRecordsTable.tsx` component
* Lifted logic for `SearchAndFilters` into `pages/LogRecords`
* Narrowed typing of common API client for log records

## Functional Testing
1. Perform testing steps as outlined in https://github.com/uwblueprint/supportive-housing/pull/74 (this will setup enough sample data locally to test the remainder of this PR)
2. Verify that the **total** number of entries is correctly displayed under the left hand side of the table
3. Verify that the **total** number of pages corresponds to the (total number of entries / entries per page) + 1
4. Verify that the number of _displayed_ records adheres to the currently selected records per page option 
5. Verify that upon changing the records per page, we are sent back to page 1
6. Verify that the displayed records change upon changing pages
7. Verify that we can manually type in a page to navigate to in the page number text field
8. Verify that the "next" and "previous" page buttons are disabled when you're on the first / last page respectively
9. Verify that the container holding the log record entries is scrollable

* Structuring of the `pages/LogRecords` ; the design choice of lifting `SearchAndFilters` was done due to weird rendering issues, but we'd still like to get a second pair of :eyes: on that
* Should we be _always_ resetting to `Page 1`  upon changing the results per page parameter?

## Checklist


frontend/src/components/common/LogRecordsTable.tsx
```diff
@@ -0,0 +1,71 @@
...
+              // TODO: Investigate alternative methods for date storage + creation
+              const dateObj = new Date(record.datetime);
+
+              return (
+                <Tr key={record.logId} style={{ verticalAlign: "middle" }}>
+                  <Td width="5%">
+                    {`${dateObj.getMonth()} ${dateObj.getDate()} ${dateObj.getFullYear()}`}
+                  </Td>
+                  <Td width="5%">
+                    {`${dateObj.getHours()}:${dateObj.getMinutes()}`}
```
safchow: on the Figma, we use 12-hour time (ex: 6:45pm) instead of 24-hour time (ex: 18:45).

frontend/src/components/common/LogRecordsTable.tsx
```diff
@@ -0,0 +1,71 @@
...
+          <Tbody>
+            {/* TODO: replace mockRecords with logRecords */}
+            {logRecords.map((record) => {
+              // TODO: Investigate alternative methods for date storage + creation
+              const dateObj = new Date(record.datetime);
+
+              return (
+                <Tr key={record.logId} style={{ verticalAlign: "middle" }}>
+                  <Td width="5%">
+                    {`${dateObj.getMonth()} ${dateObj.getDate()} ${dateObj.getFullYear()}`}
```
safchow: Would it be possible to show an abbreviated form for the month instead of the number? For example, instead of "9 2 2023", we would see Sept 9th, 2023.

frontend/src/components/common/LogRecordsTable.tsx
```diff
@@ -0,0 +1,71 @@
...
+                    {`${dateObj.getHours()}:${dateObj.getMinutes()}`}
+                  </Td>
+                  {
+                    // TODO: Resolve the resident record at some point
+                  }
+                  <Td width="5%">{record.residentFirstName}</Td>
+                  <Td whiteSpace="normal" width="75%">
+                    {record.note}
+                  </Td>
+                  <Td width="5%">{record.employeeFirstName}</Td>
```
safchow: super weird that in the designs we only use first names here, but we should be displaying both since there might be a case where two employees have the first name.

frontend/src/components/common/Pagination.tsx
```diff
@@ -0,0 +1,134 @@
+import React from "react";
```
safchow: super duper happy that this is its own component :) we can (hopefully) re-use it with a little bit of tweaking for the other pages

frontend/src/components/common/LogRecordsTable.tsx
```diff
@@ -0,0 +1,71 @@
...
+
+const LogRecordsTable = ({ logRecords }: Props): React.ReactElement => {
+  return (
+    <div>
+      <TableContainer marginTop="12px" height="70vh" overflowY="scroll">
+        <Table
+          variant="simple"
+          style={{ minHeight: "400px", verticalAlign: "middle" }}
+        >
+          <Thead className="table-header">
```
safchow: the width of the headers seem to change when flipping through pages, looks like the length of the resident name is increasing the width.
image
image

frontend/src/components/common/LogRecordsTable.tsx
```diff
@@ -0,0 +1,71 @@
...
+import { LogRecord } from "../../types/LogRecordTypes";
+
+type Props = {
+  logRecords: LogRecord[];
+};
+
+const LogRecordsTable = ({ logRecords }: Props): React.ReactElement => {
+  return (
+    <div>
+      <TableContainer marginTop="12px" height="70vh" overflowY="scroll">
```
safchow: on my screen, seems like there's a large gap when the user selects 10 items:
image

frontend/src/components/common/Pagination.tsx
```diff
@@ -0,0 +1,134 @@
...
+              justifyContent="space-evenly"
+              padding="0px 25px 0px"
+              width="210px"
+            >
+              <Text>Page</Text>
+              <NumberInput
+                maxW="60px"
+                value={pageNum}
+                size="sm"
+                onChange={handleNumberInputChange}
```
safchow, line 100: not a big fan of how the user can't press backspace on this input... here's my suggestion of how it should behave:
- user can backspace the number in the input. No number should display in the input, however, the log records will remain the same (it'll still be on the same page)
- When the user types in a new number, it'll switch to that page
- If the user clicks away without inputting a new page number, it'll reset to the original page

backend/app/services/implementations/log_records_service.py
```diff
@@ -98,9 +98,9 @@ def filter_by_flagged(self, flagged):
         print(flagged)
         return f"\nflagged={bool(flagged)}"
 
-    def get_log_records(self, page_number, filters=None):
+    def get_log_records(self, page_number, results_per_page=10, filters=None):
         try:
-            results_per_page = int(os.getenv("RESULTS_PER_PAGE"))
+            # results_per_page = int(os.getenv("RESULTS_PER_PAGE"))
```
safchow: delete unused code :D

frontend/src/components/common/Pagination.tsx
```diff
@@ -0,0 +1,134 @@
...
+                rightIcon={<ChevronDownIcon color="#6D8788" />}
+              >
+                {resultsPerPage}
+              </MenuButton>
+              <MenuList>
+                {RESULTS_PER_PAGE_OPTIONS.map((opt, ind) => (
+                  <MenuItem
+                    key={ind}
+                    onClick={() => {
+                      // TO-DO - Confirm that resetting to page 1 is what's ideal
```
safchow, line 134: so idk what the ideal behaviour is 😅 we can do two things:
- reset to page 1, ez
- figure out which page the user should be directed to. For example, page 4 with 25 results will display the same logs from page 2 with 50 results. 

Either is fine with me, but we should pick the solution that is more user-friendly

frontend/src/components/pages/LogRecords.tsx
```diff
@@ -1,106 +1,281 @@
-import React, { useState } from "react";
```
safchow, line 1: At this point, this file can probably be renamed to `Homepage.tsx`. Also, I'd create a new dir in /pages called "Homepage" and move this file and the `LogRecordsTable.tsx` file into the new dir. 

For me, the common dir is for reusable components, and it makes more sense to group components that are used once together in their own dir. 

That's my thought process, I don't want devs that are unfamiliar with this code to go "huh??? why are there two log record files". What are your thoughts on this?

frontend/src/components/pages/LogRecords.tsx
```diff
@@ -1,106 +1,281 @@
...
-                    <Td width="5%">{record.Employee}</Td>
-                    <Td width="5%">{record.Attn_To}</Td>
-                  </Tr>
-                );
-              })}
-            </Tbody>
-          </Table>
-        </TableContainer>
+        {/* Filtering dropdown */}
+        <Card>
```
safchow: oof, I'm not a big fan of removing SearchAndFilters here, but I think that it's hard to pass the states around to different components. What were the rendering issues you were running into?

frontend/src/index.css
```diff
@@ -22,7 +22,7 @@ code {
   text-align: center;
   color: #2d3748;
   width: 75%;
-  padding-top: 10%;
+  padding-top: 2%;
```
safchow: this probably needs to get updated in #76 when that's merged

frontend/src/components/common/Pagination.tsx
```diff
@@ -0,0 +1,134 @@
...
+
+  const getNumRecordsStr = () => {
+    return `${numRecords} ${numRecords === 1 ? "item" : "items"}`;
+  };
+
+  return (
+    <Box padding="12px 0px 33px">
+      <Flex justifyContent="space-between" alignItems="center">
+        <Box>
+          <Text color="#6D8788">{getNumRecordsStr()}</Text>
```
safchow: some manual styling will need to be updated when #76 is merged

safchow (changes requested):
awesome stuff 💯 you guys popped off with this and got it done quickly, but there are a few front-end issues that should be fixed so #try-again #skill-issue #be-happy

backend/app/services/interfaces/log_records_service.py
```diff
@@ -28,7 +28,7 @@ def add_record(self, log_record):
         pass
 
     @abstractmethod
-    def get_log_records(self, page_number, filters=None):
+    def get_log_records(self, page_number, results_per_page, filters=None):
```
danielk1345: If we have the default value for the results_per_page =1 in the implementations, maybe we should consider doing it the same for here

frontend/src/components/common/LogRecordsTable.tsx
kevin-pierce: Resolved 🕐

frontend/src/components/common/LogRecordsTable.tsx
kevin-pierce: Resolved ; used the following const array for the months. Let me know if you want any of these to be changed:

```
const MONTHS = ['Jan', 'Feb', 'Mar', 'Apr', 'May', 'Jun',
'Jul', 'Aug', 'Sep', 'Oct', 'Nov', 'Dec']
```

backend/app/services/implementations/log_records_service.py
kevin-pierce: image

kevin-pierce: jk done

frontend/src/components/common/LogRecordsTable.tsx
kevin-pierce: Following up on this ; the year is **not** included in the designs ; this was accidentally included by us. On that note though, is the year something we want to include?

safchow: yeah, might be confusing if you have two logs that are both jan 18th (but one is 2023, the other 2024)

kevin-pierce:
TODOs:
- Fix weird behaviour with page number text input
- Add "Day Logs" header to table
- Scrollbar should be reset to the top of the container when navigating to new page

frontend/src/components/common/LogRecordsTable.tsx
kevin-pierce: Resolved

frontend/src/components/common/LogRecordsTable.tsx
kevin-pierce: This issue will be dealt with in https://github.com/uwblueprint/supportive-housing/issues/97 ; resolving

frontend/src/components/common/LogRecordsTable.tsx
kevin-pierce: (Presumably) will be dealt with in https://github.com/uwblueprint/supportive-housing/issues/97 / another issue that will deal with screen size responsiveness ; will resolve for now

frontend/src/components/pages/LogRecords.tsx
kevin-pierce: will do

frontend/src/components/common/LogRecordsTable.tsx
kevin-pierce: will do

safchow:
@connor-bechthold @kevin-pierce can you folks also test the performance of the pagination? I noticed there was a slight delay when navigating to another page, but I don't have any issues with this unless the delay is notably longer with more log records. Please test to see what the delay is with like 10k log records and lmk 🐱

frontend/src/components/common/LogRecordsTable.tsx
safchow: the months are correct, but the year isn't showing: 
image

frontend/src/components/common/Pagination.tsx
safchow: noticing a bug here, steps to re-create:
1. Select "Show 10"
2. Go to page 5
3. Select "Show 25"
4. The input is outlined red, and says its on page 5 of 2. 
5. Clicking into the input and clicking out will navigate to page 2.

My guess is that it resets to page 1, but it doesn't update the actual page number. 
image

danielk1345: I'm also getting this error after recreating these steps!

danielk1345 (changes requested):
Not too many added comments since I think Safwaan covered most of them, but overall good work! I think the main changes and/or things to think about are the ones that Safwaan commented on (and the bug!)

frontend/src/components/common/Pagination.tsx
connor-bechthold: fixed 🙏

backend/app/services/interfaces/log_records_service.py
connor-bechthold: Good catch, will add 🙏

frontend/src/components/pages/LogRecords.tsx
kevin-pierce: Resolved ; we now have a `/HomePage` component within the `/pages` directory, and within that we have:
- HomePage.tsx (main entrypoint)
- LogRecordsTable.tsx (table component)
- SearchAndFilters.tsx (filtering component)

frontend/src/components/pages/LogRecords.tsx
kevin-pierce: Resolved as per reply in suggestion above ^

frontend/src/components/common/Pagination.tsx
```diff
@@ -0,0 +1,155 @@
...
+            />
+            <Flex
+              alignItems="center"
+              justifyContent="space-evenly"
+              padding="0px 25px 0px"
+              width="210px"
+            >
+              <Text>Page</Text>
+              <NumberInput
+                maxW="60px"
```
safchow, line 96: an issue with this is if there are 1000s of pages, the input box only shows 2 digits. This may be confusing to the user. I don't know what the width should be, but my (naïve) solution would be just to use a width that accommodates the largest possible integer. 

Syncing with the designers on this would be a good first step.

frontend/src/components/common/Pagination.tsx
```diff
@@ -0,0 +1,155 @@
...
+              onClick={() => handlePageArrowPress(userPageNum - 1)}
+            />
+            <Flex
+              alignItems="center"
+              justifyContent="space-evenly"
+              padding="0px 25px 0px"
+              width="210px"
+            >
+              <Text>Page</Text>
+              <NumberInput
```
safchow, line 95: noticing if you keep typing numbers it'll go to infinity then back to 0 lol

safchow: maybe lets cap the input to the max number of pages (meaning if there are 400 pages and I'm at 300, I add another digit, it'll just go to 400 and won't let me add new numbers)

frontend/src/components/common/Pagination.tsx
```diff
@@ -0,0 +1,155 @@
...
+    newUserPageNumString: string,
+    newUserPageNum: number,
+  ) => {
+    setUserPageNum(newUserPageNum);
+    if (
+      newUserPageNum !== pageNum &&
+      newUserPageNum >= 1 &&
+      newUserPageNum <= numPages
+    ) {
+      getLogRecords(newUserPageNum);
```
safchow: one issue I have with getting the new logs whenever the input changes is that the pagination gets really slow (I'm testing with 100000 logs) and its hard to tell when you're on the right page. I think it makes more sense to get the log records once the input is actually submitted. 

For example, if I start on page 1, and want to go to page 300, it'll fetch the logs for page 3, then page 30, then finally page 300. It should only retrieve the page 300 log records at the end, either by hitting enter on the input or clicking out of it. 

A lot of this headache is coming from the designs 😖 maybe we can approach the UI for this differently

frontend/src/index.css
```diff
@@ -14,6 +14,15 @@ code {
     monospace;
 }
 
+.records-hero {
```
safchow: #76 is merged, so merge main and move these changes to the theme dir

safchow (changes requested):
just some issues with the input for the page number, let's sync with the designers to iron them out

safchow (approved):
lgtm! there's a few thing we need to improve (dealing with large page numbers, when to call the API, etc) but I'm happy with merging it in :D 

Awesome work guys 🚀 now write 300 jest tests on pagination.

danielk1345 (approved):
LGTM! There's a few small things with the linter and whatnot but it looks good to me!

Issue #61, opened 2023-04-06
APIs for Tags Table


safchow:
## Summary
- in #60, we need to create a table for storing tags
- We also need to create the APIs for the tags
## Goals
- [ ] POST API- create tags, should be restricted to admins only. 
- [ ] GET API- get all tags, there's should be a filter by status.
- [ ] PUT API- Tags should be editable, restricted to admins only.
- [ ] PUT (DELETE) API- the delete API should really just be a put API that updates the status of the selected tag to "Deleted". This is because actually deleting a tag that is attached to log_records will cause errors if we don't first remove the tag from the log records, so we can still store the tag but just hide it from the user.
## Implementation
- Refer to existing APIs in codebase for implementation. 
- For the delete API, don't worry about removing the tag from the log record itself, we can render the tag on the front-end depending on the status of the log. 
- Blocked by #63.

PR #84, opened 2023-06-04
Linking tags and log records table

### LINKED ISSUE #63: Linking Tags and Log Records Table
## Summary
- Log Records should have a column to store the list of tags linked to that specific record.
## Goals
- [ ] Tags exists as a column in Log Records
- [ ] Tags stores a list of tags, is nullable

## Implementation
- Set the type of the column to a list of strings, we can check how to update the list in #61

connor-bechthold:
## GitHub Issue link

[Linking Tags and Log Records Table](https://github.com/uwblueprint/supportive-housing/issues/63)

## Implementation description
* Tags to log records to a many-to-many relationship, so a junction table named `log_record_tag` was used to link the existing `log_records` table and new `tags` table
* As a result of this, the create, read, update, and delete log records route services had to be updated to correctly fill in the new tables appropriately

### Setup
- Run `flask db upgrade` on the branch to ensure your local environment includes the new migrations
- You first need to populate the tags DB. Since we don't have API's for that yet, run `docker exec -it SHOW-backend /bin/bash -c "flask db upgrade"` and insert the following tags:
    - `INSERT into tags (name, status) VALUES ('tagA', 'Active');`
    - `INSERT into tags (name, status) VALUES ('tagB, 'Active');`
    - `INSERT into tags (name, status) VALUES ('tagC', 'Active');`

### Create log record functionality ###
- Test if creating a log record with an invalid tag return an error. POST `localhost:8080/log_records` with:
`{
    "attn_to": 1,
    "building": "Building A",
    "datetime": "2023-03-25 08:34:56-04:00",
    "employee_id": 1,
    "flagged": false,
    "note": "This is a note",
    "resident_first_name": "Saf",
    "resident_last_name": "Waan",
    "tags": ["wrong"]
}`
- In the DB container `select * from log_record_tag;` and `select * from log_records;` to ensure nothing was made
- Use the same body and make 3 valid log records. Use `"tags": ["tagA"]`, `"tags": ["tagA", "tagB"]`, and `"tags": ["tagA", "tagB", "tagC"]`. Once again, check the two tables in the previous step after every insert to ensure everything is being updated properly

### Get log records functionality
- Hit the GET log records route with the following (you can test more if you really really really wanna):
    - `http://localhost:8080/log_records?filters={"tags":["tagA"]}` should give you all three logs
    - `http://localhost:8080/log_records?filters={"tags":["tagB"]}` should give you two logs
    - `http://localhost:8080/log_records?filters={"tags":["tagC"]}` should give you one log
    - `http://localhost:8080/log_records?filters={"tags":["safwaan"]}` should give you no logs
    - `http://localhost:8080/log_records?filters={"tags":["tagB", "tagC"]}` should give you two logs

### Update log records functionality
- Hit the PUT log records route (which is `http://localhost:8080/log_records/{id}` where id is the id of your log record you want to update). For the body just send along the same one you used in the create section and change the tags as follows (after every request, `select * from log_record_tag;` to make sure things were updated properly ):
    - Setting tags to [] should result in the `log_record_tag` table to be empty for that log ID
    - Setting tags to ["safwaan"] should toss an error, and the table should remain unchanged
    - Any combo of `tagA`, `tagB`, and `tagC` should be valid

### Delete log records functionality
- Hitting the DELETE log records route (which is `http://localhost:8080/log_records/{id}` where id is the id of your log record you want to update). Once this is successful, the corresponding entries in the `log_record_tag` table should also be deleted

* One thing I noticed is in the relation between log records, we aren't using the ORM properly (should be using this kind of syntax in the models: `log_records = db.relationship("LogRecords", secondary="log_record_tag", back_populates="tags")`. We should probably look into updating this eventually as it makes managing both tables a lot less painful
* The SQL query we're using when getting log records is becoming hella large, and it's probably super slow. With an ORM, there shouldn't be a need to even use raw SQL in the first place. We should look into changing this to only use SQLAlchemy queries (from the brief investigation I did, this should be possible)

## Checklist


backend/app/models/log_record_tag.py
```diff
@@ -0,0 +1,32 @@
+from sqlalchemy import inspect
+from sqlalchemy.orm.properties import ColumnProperty
+
+from . import db
+
+
+class LogRecordTag(db.Model):
+    __tablename__ = "log_record_tag"
+
+    log_record_tag_id = db.Column(db.Integer, primary_key=True, nullable=False)
```
safchow: for consistency, I mentioned we should either do <table_name>_id or just id. I'm leaning towards switching everything to id. What are your thoughts?

The only reason for this switch is because the resident table has id and resident_id which is a little confusing at first.

backend/app/models/tag.py
```diff
@@ -12,6 +12,7 @@ class Tag(db.Model):
     status = db.Column(
```
safchow: thinking this filename should be tags instead of tag (all of our models use plural instead of singular except user, which should also be updated)

safchow: should also be applied to the log_record_tags file

backend/app/models/tag.py
safchow, line 1: Realizing this now but name should probably be unique. There should be some validation (both in the database as a table constraint and the routes layer to validate that the name is unique).

safchow, line 1: Also, maybe status isn't the best approach... it's creating unused data in the db which isn't a big deal in the grand scheme of things but its not ideal...

what if we remove the status column entirely, and do a delete similar to how log_records are handled? First, delete all the occurences of tag_id in the new table, log_record_tag, then delete the tag from its table? This would make sure we don't have unused data in the db and makes it easier on the user (we would've needed to show unused tags to the user as well)

safchow: I think the one complaint about this approach was that any log record with a specific tag would lose that tag, but we can throw a warning to the admin and indicate to them that every log with the selected tag will lose that tag

safchow: this new approach would need to be handled on the API side so it doesn't need to be addressed here but we should have something concrete for #61

backend/app/services/implementations/log_records_service.py
safchow, line 1: noticing some goofy stuff on line 62-63, the indexes are swapped for tags and building. Here is a default response, note that building and tags are swapped:
```
{
    "log_records": [
        {
            "attn_to": 1,
            "attn_to_first_name": "Safwaan",
            "attn_to_last_name": "Chowdhury",
            "building": [
                "tagA",
                "tagB"
            ],
            "datetime": "2023-06-13 15:04:12.437682-04:00",
            "employee_first_name": "Safwaan",
            "employee_id": 1,
            "employee_last_name": "Chowdhury",
            "flagged": false,
            "log_id": 7,
            "note": "This is a note",
            "resident_first_name": "Saf",
            "resident_last_name": "Waan",
            "tags": "Building D"
        }
    ],
    "num_results": 1
}
```

Swapping the indexes of tags and building, note that building is index 8, and tags is index 9:
```
                        "building": log[8],
                        "tags": log[9],
```

fixes this issue:
```
{
    "log_records": [
        {
            "attn_to": 1,
            "attn_to_first_name": "Safwaan",
            "attn_to_last_name": "Chowdhury",
            "building": "Building D",
            "datetime": "2023-06-13 15:04:12.437682-04:00",
            "employee_first_name": "Safwaan",
            "employee_id": 1,
            "employee_last_name": "Chowdhury",
            "flagged": false,
            "log_id": 7,
            "note": "This is a note",
            "resident_first_name": "Saf",
            "resident_last_name": "Waan",
            "tags": [
                "tagA",
                "tagB"
            ]
        }
    ],
    "num_results": 1
}
```

backend/app/services/implementations/log_records_service.py
```diff
@@ -25,13 +25,26 @@ def add_record(self, log_record):
         new_log_record = log_record
         new_log_record["datetime"] = datetime.now()
 
+        tag_names = new_log_record["tags"]
```
safchow, line 28: not sure what specifically needs to change, but the response after creating a log record doesn't have the tags:
```
{
    "attn_to": 1,
    "building": "Building D",
    "datetime": "Tue, 13 Jun 2023 19:15:40 GMT",
    "employee_id": 1,
    "flagged": false,
    "note": "This is a note",
    "resident_first_name": "Saf",
    "resident_last_name": "Waan"
}
```

backend/app/services/implementations/log_records_service.py
```diff
@@ -86,9 +99,9 @@ def filter_by_date_range(self, date_range):
         return f"\ndatetime>='{start_date}' AND datetime<='{end_date}'"
 
     def filter_by_tags(self, tags):
-        sql_statement = f"\n'{tags[0]}'=ANY (tags)"
+        sql_statement = f"\n'{tags[0]}'=ANY (tag_names)"
```
safchow: when filtering, I'm noticing log records appear when they shouldn't. In this example, this is the query param: `filters={"tags":["tagA", "TagB"]}`

Here is the response:
```
{
    "log_records": [
        {
            "attn_to": 1,
            "attn_to_first_name": "Safwaan",
            "attn_to_last_name": "Chowdhury",
            "building": "Building D",
            "datetime": "2023-06-13 15:15:40.849777-04:00",
            "employee_first_name": "Safwaan",
            "employee_id": 1,
            "employee_last_name": "Chowdhury",
            "flagged": false,
            "log_id": 8,
            "note": "This is a note",
            "resident_first_name": "Saf",
            "resident_last_name": "Waan",
            "tags": [
                "tagA"
            ]
        },
        {
            "attn_to": 1,
            "attn_to_first_name": "Safwaan",
            "attn_to_last_name": "Chowdhury",
            "building": "Building D",
            "datetime": "2023-06-13 15:04:12.437682-04:00",
            "employee_first_name": "Safwaan",
            "employee_id": 1,
            "employee_last_name": "Chowdhury",
            "flagged": false,
            "log_id": 7,
            "note": "This is a note",
            "resident_first_name": "Saf",
            "resident_last_name": "Waan",
            "tags": [
                "tagA",
                "tagB"
            ]
        }
    ],
    "num_results": 2
}
```

The first log record shouldn't appear, yes it does have tag A, but it doesn't have tag B. We should only show log records that have both tag A and tag B, and other tags if any. For example, suppose another log record existed with A, B, C- that log record should appear if A and B are filtered for. 

Sorry if I explained it poorly the first time 😖

safchow (changes requested):
Senior dev moment 😤 coming up with junction tables actually popped off MASSIVELY

There is some changes I suggested and some things to fix #skill-issue #junior-developer-moment #get-good #lbtm

connor-bechthold:
Future comment - investigate use of JSON_AGG for grouping tag row aggregation

safchow (approved):

danielk1345 (approved):

PR #96, opened 2023-06-05
User and Residents Seeding Script

safchow:
## Implementation description
* Working off of the changes in #74, I created a bash script to seed your own user into the DB. This was done because every time your database got reset, you had to delete your Firebase user and sign up again which was super tedious. With this script, you don't have to delete your Firebase user on a fresh DB. 
* I also added a script to seed Residents (since #50 was merged recently)

### Setup
Add the following to your /seeding/.env file:
  ```
  # Residents Table
  RESIDENT_ROWS=5

  #User
  FIRST_NAME='<your first name>'
  LAST_NAME='<your last name>'
  AUTH_ID='<your auth id, can be found in the local DB>'
  ROLE='Admin'
  ```

### Testing
- Delete your user from the db (after you get your auth_id) and try to log in, verify you get a 500
- Run the script then try to log in again. Validate you can successfully log in
- Verify the `create-residents.sh` script runs correctly
- Verify the `create-log-records.sh` script is also working, since that was tweaked


seeding/create-log-records.sh
```diff
@@ -0,0 +1,36 @@
...
+  attn_to, 
+  note, 
+  tags, 
+  building 
+) 
+SELECT
+  $EMPLOYEE_ID,
+  (ARRAY['John','Jerry','James','Jack'])[floor(random() * 4 + 1)], 
+  (ARRAY['Doe','Show','Flow','Mow'])[floor(random() * 4 + 1)], 
+  NOW() + (random() * (NOW()+'90 days' - NOW())) + '30 days',
```
Safewaan, line 26: @connor-bechthold this isn't the right date format for the `GET` API, throws this error when called:
```
{
    "error": "'bool' object has no attribute 'astimezone'"
}
```

seeding/seed.sh
```diff
@@ -1,12 +1,9 @@
 #!/bin/bash
 
-#If you're on Windows, run bash seed.sh -w
-#Otherwise, run bash seed.sh
-
-#Import root env file variables
+# Import root env file variables
```
kevin-pierce: Just to improve clarity

```suggestion
# Import /seeding directory env file variables
```

seeding/create-user.sh
```diff
@@ -0,0 +1,30 @@
...
+  '$LAST_NAME',
+  '$AUTH_ID',
+  '$ROLE'
+WHERE NOT EXISTS (
+  SELECT 1
+  FROM users
+  WHERE first_name = '$FIRST_NAME'
+    AND last_name = '$LAST_NAME'
+);
+"
```
connor-bechthold, line 29: If I'm correct here, this inserts the user you want to add only if a user with the same first and last name doesn't exist? I'm probably missing the point of this, but multiple users can have the same name no?

seeding/create-log-records.sh
```diff
@@ -0,0 +1,38 @@
...
+  datetime, 
+  flagged, 
+  attn_to, 
+  note, 
+  tags, 
+  building 
+) 
+SELECT
+  $EMPLOYEE_ID,
+  (ARRAY(SELECT generate_series(1, $RESIDENT_ROWS)))[floor(random() * $RESIDENT_ROWS + 1)],
```
connor-bechthold, line 28: This works, but I could see this becoming annoying when for ex. 
- Insert 25 residents, they have ID's 1 through 25
- Delete them
- Insert 25 more residents, they now have ID's 26 through 50
- There's 25 resident rows, but now the ID's are out of range

Kind of an edge case to be honest, but there's probably less room for error to tie it to a specific value. It would be even better if we could get the current IDs from the DB.

If we're going this route, I think it would be nice to expand this to EMPLOYEE_ID and ATTN_TO_ID as well

connor-bechthold (changes requested):
Looks dope, added a few comments to look at

For fixing the path issues, add this to the top of every .sh file with your imports
```
SEEDING_DIR=$( cd -- "$( dirname -- "${BASH_SOURCE[0]}" )" &> /dev/null && pwd )
source $SEEDING_DIR/seed.sh <-- Example for importing seed.sh, you can do the same for .env
```

seeding/create-user.sh
safchow: this is just a check to make sure you're not inserting yourself into the db if you already exist

safchow: also you're right, multiple users can have the same name (like Jonathon Jones) but since this is a script file for us only it should be okay

seeding/create-log-records.sh
safchow: yeah you're right :( I'm thinking the auto_increment can be reset in the script before running the insert again

safchow: Added a fix for this, here's the queries in the script:
1. Delete all the residents in the database.
2. Reset the id sequence back to 1.
3. Insert the specified number of residents into the table.

This should work and resolve the error you mentioned.

connor-bechthold (approved):
lgtm 🥵

kevin-pierce (approved):
🐐

PR #105, opened 2023-06-18
Adding styling for new log components

### LINKED ISSUE #82: Create Log Styling
## Summary

The Create Log must be styled according to the Figma designs. This task is blocked until #76 is merged. 
react-select documentation: https://react-select.com/home

## Goals

- [ ] Add icons to the next of the inputs. Icons can be found [here](https://chakra-ui.com/docs/components/icon)
- [ ] Add styling to all the inputs (selects, text input, checkbox)

## Implementation

- Refer to how styling is done in the design system.

phamkelly17:
## GitHub Issue link

[Create Log Styling](https://github.com/uwblueprint/supportive-housing/issues/82)

## Implementation description
* created styling for select, single date picker, and added necessary styling for other components
* NOTE: since we used react-select library instead of using Chakra's select component, we can not override the style like how we do it using the design system for other components (i.e. button)

Screen Shot 2023-06-18 at 6 03 58 PM
Screen Shot 2023-06-18 at 6 03 51 PM
Screen Shot 2023-06-18 at 6 03 42 PM
Screen Shot 2023-06-18 at 6 03 28 PM
Screen Shot 2023-06-18 at 6 03 20 PM

* I can not for the life of me figure out how to change the colors on the time picker. It is a ChakraUI Input component with `type="time"` and I'm not entirely convinced that it can be changed at all (ChatGPT failed me)
Screen Shot 2023-06-18 at 6 03 34 PM

## Checklist


frontend/src/theme/selectStyle.tsx
```diff
@@ -0,0 +1,55 @@
+import colors from "./colors";
+
+const optionBackgroundColor = (isSelected, isFocused) => {
```
safchow: seeing this error when I compile:
```
/app/src/theme/selectStyle.tsx
TypeScript error in /app/src/theme/selectStyle.tsx(3,32):
Parameter 'isSelected' implicitly has an 'any' type.  TS7006

    1 | import colors from "./colors";
    2 | 
  > 3 | const optionBackgroundColor = (isSelected, isFocused) => {
      |                                ^
    4 |   if (isSelected) {
    5 |     return colors.teal[100];
    6 |   }
```

safchow: setting their types to boolean doesn't fix this :(
```
/app/src/theme/selectStyle.tsx
TypeScript error in /app/src/theme/selectStyle.tsx(3,32):
Parameter 'isSelected' implicitly has an 'any' type.  TS7006

    1 | import colors from "./colors";
    2 | 
  > 3 | const optionBackgroundColor = (isSelected: boolean, isFocused: boolean) => {
      |                                ^
    4 |   if (isSelected) {
    5 |     return colors.teal[100];
    6 |   }
```

frontend/src/theme/colors.tsx
```diff
@@ -17,7 +17,7 @@ const colors = {
   },
   teal: {
     50: "#EAF0EF",
-    100: "#285E61",
+    100: "#C6CFD2",
```
safchow: good catch- seems like the label on Figma and the actual colour aren't the same

frontend/src/theme/singleDatePickerStyles.tsx
```diff
@@ -0,0 +1,44 @@
+import colors from "./colors";
```
safchow: woah you added styling for the date picker 😄 it wasn't part of the PR but it's definitely helpful to have for the future. Unfortunately, we don't have designs for a single date picker so I'll add that to the to-do list for the designers

safchow:
> * I can not for the life of me figure out how to change the colors on the time picker. It is a ChakraUI Input component with `type="time"` and I'm not entirely convinced that it can be changed at all (ChatGPT failed me)

So the number wheel thing for dates is actually something that's coming from Chrome. On Mozilla:
image

On Chrome:
image

We can just hide the icon as a workaround, this should help: https://stackoverflow.com/questions/61934148/hide-the-icon-on-a-html-time-input-field-in-chrome

safchow (commented):
I wasn't able to fully test because of a compile error 🥲 lmk if you're also seeing it

frontend/src/theme/selectStyle.tsx
phamkelly17: Mine works fine with this change, but then again it also worked when no type is specified

frontend/src/theme/singleDatePickerStyles.tsx
phamkelly17: i just changed the colors so they would be purple and removed the ugly blue outline that appears when you click on it :D

frontend/src/theme/selectStyle.tsx
kevin-pierce: EDIT:

@Safewaan - Just double check that you hit save / rebuild the app ; I implemented that fix and it was fine

phamkelly17: @kevin-pierce could you try adding the change safwaan tried two comments above because that wasn't working for him either

kevin-pierce: Yeah it worked locally, though another issue arose after implementing that fix (don't know if I'm missing some step in branch setup):
Screen Shot 2023-06-19 at 8 54 46 PM

phamkelly17: no I added that T_T... could you just try changing that to `any` type

kevin-pierce: 🎉 

Screen Shot 2023-06-19 at 9 00 04 PM

frontend/src/theme/colors.tsx
```diff
@@ -17,7 +17,7 @@ const colors = {
   },
```
kevin-pierce: nit: I don't wanna be that one person (but I'm gonna be that one person) - can we rename all references from `colors` to `colours` 🤓

kevin-pierce: Actually wait this will just unnecessarily add a bunch of diffs, I take this back

frontend/src/theme/selectStyle.tsx
```diff
@@ -0,0 +1,55 @@
...
+  control: (provided: any, state: any) => ({
+    ...provided,
+    backgroundColor: state.isDisabled ? colors.teal[50] : colors.white.default,
+    borderColor: colors.gray[200],
+    boxShadow: "none",
+    borderRadius: "4px",
+    "&:hover": {
+      border: `1px solid ${colors.gray[200]}`,
+      cursor: "pointer",
+    },
```
kevin-pierce: nit: The same way we extracted out the `optionBackgroundColor` function, could we extract out the `multiValue`, `singleValue`, `control`, and `option` methods? Will make this file a bit more consistent

kevin-pierce:
```suggestion
  const controlStyle = (provided: any, state: any) => ({
    ...provided,
    backgroundColor: state.isDisabled ? colors.teal[50] : colors.white.default,
    borderColor: colors.gray[200],
    boxShadow: "none",
    borderRadius: "4px",
    "&:hover": {
      border: `1px solid ${colors.gray[200]}`,
      cursor: "pointer",
    },
    })

  control: controlStyle(provided, state),
```

frontend/src/components/forms/CreateLog.tsx
```diff
@@ -331,14 +322,17 @@ const CreateLog = () => {
                       placeholder="Building No."
                       onChange={handleBuildingChange}
                       styles={{
```
safchow: The changes in this file are a combination of changes, so I'll add the explanation for the main changes in the `selectStyle.tsx` file. Update the styles to only use selectStyle
```suggestion
                      styles={selectStyle}
```

frontend/src/components/forms/CreateLog.tsx
```diff
@@ -353,14 +347,17 @@ const CreateLog = () => {
                       placeholder="Select Resident"
                       onChange={handleResidentChange}
                       styles={{
```
safchow: Update the styles to only use selectStyle
```suggestion
                      styles={selectStyle}
```

frontend/src/components/forms/CreateLog.tsx
```diff
@@ -389,6 +387,18 @@ const CreateLog = () => {
                       options={EMPLOYEES}
                       placeholder="Select Employee"
                       onChange={handleAttnToChange}
+                      styles={{
```
safchow: Update the styles to only use selectStyle
```suggestion
                      styles={selectStyle}
```

frontend/src/theme/selectStyle.tsx
```diff
@@ -0,0 +1,55 @@
...
+  singleValue: (provided: any) => ({
+    ...provided,
+    color: colors.gray[500],
+    borderColor: colors.gray[200],
+    boxShadow: "none",
+  }),
+  control: (provided: any, state: any) => ({
+    ...provided,
+    backgroundColor: state.isDisabled ? colors.teal[50] : colors.white.default,
+    borderColor: colors.gray[200],
```
safchow: When the select is focused, it should update to teal.400. Otherwise, it should default to teal.100:
```suggestion
    borderColor: state.isFocused ? colors.teal[400] : colors.teal[100],
```

one issue with the code above is that after you make your selection, the state will still remain focused until you click elsewhere. I don't think its too much of a big deal, since it technically shows the user their last selection/action

frontend/src/theme/selectStyle.tsx
```diff
@@ -0,0 +1,55 @@
...
+    boxShadow: "none",
+    borderRadius: "4px",
+    "&:hover": {
+      border: `1px solid ${colors.gray[200]}`,
```
safchow: it was kind of hard to track down what should happen when you hover, but I think the new color should be teal.300. 
```suggestion
      border: `1px solid ${colors.teal[300]}`,
```

frontend/src/theme/selectStyle.tsx
```diff
@@ -0,0 +1,55 @@
...
+  if (isFocused) {
+    return colors.teal[50];
+  }
+  return colors.white.default;
+};
+
+const selectStyle: any = {
+  singleValue: (provided: any) => ({
+    ...provided,
+    color: colors.gray[500],
```
safchow: the default color text should be teal.300
```suggestion
    color: colors.teal[300],
```

I don't know where to make the same change so that the tags select uses teal.300 😅 

Also, when the select is focused, the color should update to teal.400. I think you have to pass state as a parameter, then do something like:
``` suggestion
    color: state.isFocused ? colors.teal[400] : colors.teal[300],
```

frontend/src/components/forms/CreateLog.tsx
safchow: oops I forgot to add the explanation but basically, we should keep all the styling in `selectStyle`. The error border rendering was removed in my suggestion because it was something I added in but it wasn't based off of any designs.

frontend/src/components/forms/CreateLog.tsx
```diff
@@ -276,44 +267,44 @@ const CreateLog = () => {
...
+                        isInvalid={timeError}
+                        style={{ color: colors.gray[500] }}
+                      >
                         <FormLabel>Time</FormLabel>
                         <Input
                           size="md"
                           type="time"
                           defaultValue={time}
                           onChange={handleTimeChange}
+                          style={{
```
safchow: there should be more styling for these inputs (such as what happens when the input is hovered, focused, etc.) 

The good thing is that its the exact same styling as the selects, so use that as a reference point

safchow (changes requested):
thanks for looking into why the code wasn't compiling, I was able to review the PR properly 😄 

I left a few comments on super specific things but overall, amazing job🔥you're on your way to becoming top contributor in June 😤😤😤

frontend/src/components/forms/CreateLog.tsx
safchow: the inputs are time, date and note. It might make sense to create a file like `formInput.tsx` and add the styling there so we can reuse it for the other forms

safchow:
the designers suggested that we add horizontal lines to distinguish different parts of the modal like so:

image

frontend/src/theme/selectStyle.tsx
phamkelly17: the thing is, the only way i could figure out how to remove the ugly blue border when you focus on the non "react-select" library components was to make the border color not change, which is why I made it this way. I'll try again but idk if i'll be able to make it work w the date time pickers

safchow (approved):
LGTM (looks garbage to me) 🧄

connor-bechthold (approved):
lgtm hopefully

PR #115, opened 2023-06-21
[FIX] - Pagination Component redundant fetches

### LINKED ISSUE #111: Pagination Component - Prevent redundant / unintentional fetches
## Summary
In order to throttle the page-search functionality on the homepage, we want to implement some form of debouncing / throttling. In the current implementation, when users search for a large page, such as `400`, the following will happen:
- User types '4' ==> fetch triggered for page 4
- User types '0' ==> fetch triggered for page 40
- User types '0' ==> fetch triggered for page 400

As per discussion, the approach of choice entails leveraging `onBlur` to check for when the user exits the textfield for the page in order to initiate a new fetch. 

## Goals

- [x] Trigger a new fetch when the user exits the textfield through `onBlur`
- [x] Trigger a new fetch when the user hits the `enter` key

kevin-pierce:
## GitHub Issue link

[Pagination Component - Prevent redundant / unintentional fetches](https://github.com/uwblueprint/supportive-housing/issues/111)

## Implementation description
* Created `fetchRecords` function to actually perform the get
* Added `onKeyUp` event handler to trigger fetch
* Modified `handleNumberInputChange` function to _only_ update the input value

## Steps to test
1. Seed log records (sufficient number such that we end up with some # of pages > 10)
2. Verify that changing the page number (without clicking outside of the text field / hitting enter) will NOT cause a refetch of log records
3. Verify that changing the page number and unfocusing the text field (whether through `tab` or clicking outside of it) will refetch log records
4. Verify that changing the page number and hitting enter will refetch the log records
5. Verify that hitting enter OR unfocusing the text field **without** changing the page number will NOT refetch log records

* The structure / call stack (might be a little confusing / messy) + there was some discussion around making this a reusable component in the future ; perhaps we might want to restructure the component within this PR itself?

## Checklist


frontend/src/components/common/Pagination.tsx
```diff
@@ -50,15 +50,22 @@ const Pagination = ({
...
-    }
   };
+  
+  const fetchRecords = () => {
+    if (!Number.isNaN(userPageNum) && userPageNum !== pageNum) {
+      getRecords(userPageNum);
+    }
+  }
 
-  const handleBlur = (event: React.FocusEvent<HTMLInputElement>) => {
```
connor-bechthold, line 58: So we actually want to keep this in here because we want to keep this behaviour:
- Clear the text field completely 
- Click out of the text field (triggering onBlur)
- The current page number is filled back into the text box

Right now, if you clear the box and click out of it, the box stays empty, which isn't what we want

connor-bechthold (changes requested):
It's crazy how much better of a user experience this is 🧑‍🍳 

Couple of things:
- I think the flow would be even better if when you click enter, you're deselected from the box
- Not sure if this is out of the scope of this ticket, but do we also want to consider disabling the inputs while the records are being fetched? cc @Safewaan

frontend/src/components/common/Pagination.tsx
kevin-pierce: Ah yeah good catch, will fix - should not have written this at 12 am clearly 💀

frontend/src/components/common/Pagination.tsx
```diff
@@ -50,15 +50,28 @@ const Pagination = ({
...
+    }
+  }
+
+  const handleBlur = () => {
     if (Number.isNaN(userPageNum)) {
       setUserPageNum(pageNum);
+      return;
     }
+    fetchRecords()
   };
```
kevin-pierce: This control flow is a little weird, so I thought I'd comment it to explain:

Two ways to trigger a fetch:
- Exiting focus from the textbox (Clicking outside of it) ; **handleBlur**
- Hitting the enter key ; **handleKeyUp**

handleBlur:
- Checks if number is present ; if not, reset to the previous number and terminate
- If a valid number is present, initiate a fetch

handleKeyUp:
- Checks if key-up (triggered once a user "stops" pressing a key) is == 13 (the key code for the `ENTER` key) ; if true, then this will force a `blur` event (triggering the `handleBlur` function)

safchow:
>     * Not sure if this is out of the scope of this ticket, but do we also want to consider disabling the inputs while the records are being fetched? cc @Safewaan

Let's see what the performance of pagination looks like after the improvements and discuss with the designers

safchow (approved):
LGTM THIS IS 🔥 (fire)

connor-bechthold (approved):
This is sickkkkk lets goo 👨‍🍳

PR #116, opened 2023-06-22
[QOL] - ESLint on save

kevin-pierce:
This is simply a minor QoL upgrade ; with these changes, ESLint should now run on-save (for anyone using VSCode). Initially, this might cause some ugly diffs in your files (due to the restructuring of imports) but hopefully this makes linting a little less annoying going forward.

## Checklist


connor-bechthold (commented):
I'm not too familiar with this but couple of things
- If we do this we should probably include all the reformatting that's done because this is going to bloat future PR's
- Also just realized, the linter that we typically run is through yarn, and not ESlint. Curious why we have both? If the yarn linter works for us now, we can probably stick with that instead of going through the hassle of reformatting things

PR #122, opened 2023-06-24
Improvement: Pagination Processing Time

### LINKED ISSUE #102: Improve Pagination Processing Time
When the database has a ton of log records (10000+), the pagination becomes noticeably slower. 

This might be a good starting point: https://stackoverflow.com/questions/59619956/pagination-and-filtering-on-a-very-large-table-in-postgresql-keyset-pagination

https://github.com/uwblueprint/supportive-housing/assets/57375371/042c55aa-84db-4103-8869-4ff769223c5c

connor-bechthold:
## GitHub Issue link

[Issue](https://github.com/uwblueprint/supportive-housing/issues/102)

### **Problem:**
Pagination was super slow when lots of records were in the database. Why? Everytime we want a new page, all the records matching the required filters were fetched, and a subset of those was returned through Python indexing (slow and inefficient)

### **Solution:**
We should only be querying what we want. In the query for log records, we now use LIMIT and OFFSET
- LIMIT: How many records we want to query for (i.e. results per page)
- OFFSET: Where we want to start getting those records from (i.e. page number)

One subsequent problem is that we couldn't return the total number of records from that query because we weren't getting every associated record anymore. Thus, that existing route strictly returns the log records. We created a separate route/query that gets the number of results for that query. Getting the total count is a slow process (nothing we can do about it), and will limit performance on the frontend.

However, the performance limitation should be understandable. Every time the filters change we make two API calls:
- One to get the new total count
- One to get the log records

This should be expected to be a bit slower since the filters changed. On page number change or results per page change, we only make one call to the newest log records. The count stays the same, so this process is much faster.

### **Side Effect**
Concurrent log record adds are still an issue. There's no good way to get a new count in realtime, so the user will have the oldest log records until they refresh. The only situation this doesn't happen is for example:
- The user's last page is only displaying 5 records out of 10 spots (10 results per page)
- If 5 records are added and the user changes the page, those records will be there when the user goes back to that page
- If another record is added, it won't be seen since it's on another page

### Demo
Here is an example of flipping through 100k logs, which is considerably faster when compared to the original vid posted in the GH issue:

https://github.com/uwblueprint/supportive-housing/assets/64047863/cf4e761a-8be6-43d4-8f94-44e4f8c7c37a

## Steps to test
1. Add a very large number of records to the database
2. Ensure flipping through pages doesn't take longer than around a second or two (internet connection could affect this)

* Should we have a separate button to trigger a count refresh? Little annoying that on an entire page refresh, the filters are reset

## Checklist


frontend/src/APIClients/CommonAPIClient.ts
```diff
@@ -1,9 +1,15 @@
...
 import baseAPIClient from "./BaseAPIClient";
 
-import { LogRecordFilters } from "../components/common/types/Filters";
+import {
+  LogRecordFilters,
+  CountLogRecordFilters,
+} from "../components/common/types/Filters";
 
 const filterLogRecords = async ({
   building = "",
```
phamkelly17, line 15: in the model, building is not nullable, so should we be setting a default of ""?

phamkelly17 (approved):
lgtm

safchow (approved):
DEMON PR 😈 

apologies for the late review, but thanks for investigating how to improve our pagination and getting it in quickly :D

PR #133, opened 2023-07-07
Invite Employees modal

### LINKED ISSUE #94: Invite Employee Form
## Summary

On the Resident page, there should be a modal to allow admins to invite an employee to the system. Here are the inputs required in the form:
- First Name
- Last Name
- Show Email- Some validation should exist to verify that this is in correct email format
- Role- Should be a radio component with two choices: a) Admin b) Non Admin
- Role (continued)- After the radio component, there should be a checkbox for 2FA

This ticket should also cover the front-end/back-end integration, meaning that once the form is submitted, the employee is correctly invited and a user is created in the DB. A similar task is being worked on for creating logs- #81. You can also refer to how the endpoints are called in the `InviteUsers` since a demo was created for inviting users, and a lot of the code can be reused from there. 

Note: the designs for this modal aren't fully complete yet, but there is enough to use to begin working on this task. For any uncertainties, either sync with me or the designers.

## Goals

- [ ] Create a new form called `InviteEmployees`
- [ ] Add all the inputs mentioned above
- [ ] Ensure form validation is present
- [ ] When the form is submitted and all inputs are valid, create a user invite in the db

## Implementation

1. To test the modal, I'd recommend just adding a separate page (maybe something like testInviteEmployee) then adding a button that opens the modal
2. A lot of the code can be reused from `CreateLog`
3. Also please reuse the API calls and integration written in `InviteUsers`, it will save a lot of time :D

helenellyx:
## GitHub Issue link

[Invite Employee Form](https://github.com/uwblueprint/supportive-housing/issues/94)

## Implementation description
*  a modal to allow admins to invite an employee to the system
* once the form is submitted, the employee is correctly invited and a user is created in the DB

Screenshot 2023-07-06 at 11 25 21 PM

## Steps to test
1. Make sure you are logged in as an admin
2. Go to http://localhost:3000/invite-employees
3. Open the modal
4. Note that these inputs are required: `First Name, Last Name, Admin Status`, so play around with it and see if errors pop up when you try to submit but some or all of those fields are empty
Screenshot 2023-07-06 at 11 32 55 PM

5. make sure that the email input field must be a SHOW email, and a warning pops up if not a SHOW email
6. make sure that when the `admin radio` is selected, the checkbox is disabled
7. Now try actually inviting the users and check that it is in the users db

- when you select: admin, the user role is admin
- when you select non-admin + needs 2Fa, the user role is relief staff
- when you select non-admin + doesn't need 2fa, the user role is regular staff

8. Make sure the success toast shows up
Screenshot 2023-07-07 at 11 31 26 AM

10. Also do some hardcode to see if the error toast shows up correctly

* the user roles are right 
* toasts
* Does the font difference between the input boxes and the radios + checkbox font matter

## Checklist


frontend/src/components/forms/InviteEmployeesModal.tsx
```diff
@@ -0,0 +1,314 @@
...
+  Radio,
+  RadioGroup,
+  Text,
+} from "@chakra-ui/react";
+import commonApiClient from "../../APIClients/CommonAPIClient";
+import RoleOptions from "../common/types/Roles";
+import { INVITE_EMPLOYEE_ERROR } from "../../constants/ErrorMessages";
+import CreateToast from "../common/Toasts";
+
+const InviteEmployeesModal = (): React.ReactElement => {
```
safchow: Can you rename this to `CreateEmployee`?

frontend/src/components/forms/InviteEmployeesModal.tsx
```diff
@@ -0,0 +1,314 @@
...
+      !invitedLastNameError &&
+      !invitedAdminStatusError
+    ) {
+      onInviteEmployee();
+    }
+  };
+
+  return (
+    <>
+      <Button size="lg" variant="solid" colorScheme="pink" onClick={handleOpen}>
```
safchow: pink button goes crazy 🪅 

I would change this to what we would use in the app, you can remove the styling and just use the primary variant

```suggestion
      <Button variant="primary" onClick={handleOpen}>
 ```
 
 and also change the button text so it has an add icon and the text is something like `Invite Employee`

File: frontend/src/components/forms/InviteEmployeesModal.tsx [hunk 3]
```diff
@@ -0,0 +1,314 @@
...
+        <ModalOverlay />
+        <ModalContent minW="513px" maxW="513px" minH="484px" borderRadius="8px">
+          <Box
+            pl="24px"
+            pr="16px"
+            mt="10px"
+            mb="10px"
+            ml="10px"
+            mr="10px"
+            borderBottom="1px solid #EBECF0"
```
safchow: I don't think we need the styling here for the box component, we didn't add it for the other modals.

frontend/src/components/forms/InviteEmployeesModal.tsx
```diff
@@ -0,0 +1,314 @@
...
+                >
+                  <Flex>
+                    <Box marginRight="24px">
+                      <Radio
+                        size="lg"
+                        colorScheme="teal"
+                        color="black"
+                        value="1"
+                      >
+                        Admin
```
safchow: I would wrap the text in a Text component, so it gets the default styling 
```suggestion
                        <Text>Admin</Text>
 ```

File: frontend/src/components/forms/InviteEmployeesModal.tsx [hunk 5]
```diff
@@ -0,0 +1,314 @@
...
+                      </Radio>
+                    </Box>
+                    <Box>
+                      <Radio size="lg" colorScheme="teal" value="2">
+                        Non Admin
```
safchow: same as above

frontend/src/components/forms/InviteEmployeesModal.tsx
```diff
@@ -0,0 +1,314 @@
...
+                size="lg"
+                colorScheme="teal"
+                borderRadius="40px"
+                isChecked={isTwoFactorAuthenticated}
+                isDisabled={isDisabled}
+                onChange={(event) =>
+                  setIsTwoFactorAuthenticated(event.target.checked)
+                }
+              >
+                Require Two Factor Authentication
```
safchow: same as above

frontend/src/components/forms/InviteEmployeesModal.tsx
```diff
@@ -0,0 +1,314 @@
...
+      setInvitedLastNameError(false);
+    } else {
+      setInvitedLastNameError(true);
+    }
+  };
+
+  const handleEmailChange = (e: { target: { value: unknown } }) => {
+    const inputValue: string = e.target.value as string;
+    setInvitedEmail(inputValue);
+    if (inputValue.endsWith("@supportivehousing.com")) {
```
safchow: I'm not sure if we should check if they have a supportive housing email, they may want to create accounts with other addresses in mind. Instead, checking if the email is in a valid format is sufficient

safchow: also also, I would move this check to the submit instead, since as the user is typing, the error validation will go off when we want it to appear when the user submits but the form is invalid

safchow (changes requested):
haven't done a full review but putting comments so far up

frontend/src/components/forms/InviteEmployeesModal.tsx
helenellyx: but in the designs, theres this line on the bottom, I think the box is good to keep for designs
Screenshot 2023-07-07 at 8 53 29 PM

safchow: we use a `Divider` component for the horizontal lines in the forms instead of a `borderBottom` prop

frontend/src/components/forms/CreateEmployee.tsx
```diff
@@ -0,0 +1,320 @@
...
+                <FormControl isRequired isInvalid={invitedEmailError}>
+                  <FormLabel>SHOW Email</FormLabel>
+                  <Input
+                    placeholder="Enter SHOW email"
+                    value={invitedEmail}
+                    onChange={(event) => setInvitedEmail(event.target.value)}
+                    maxLength={254}
+                  />
+                  <FormErrorMessage>
+                    A validwh email is required.
```
safchow: typo here 🌵

frontend/src/components/forms/CreateEmployee.tsx
```diff
@@ -0,0 +1,320 @@
...
+      if (hasInvitedUser) {
+        newToast(
+          "Invite sent",
+          `Your invite has been sent to ${invitedFirstName} ${invitedLastName}`,
+          "success",
+        );
+        handleClose();
+      }
+    } else {
+      newToast("Error inviting employee", INVITE_EMPLOYEE_ERROR, "error");
```
safchow: we don't need to show a toast message here because the form validation is enough to indicate to the user that something is invalid BUT its good practice to add error messages when needed :')

Some better use cases of displaying an error toast message is if the form is valid but somehow the submission failed (api returns an error)

safchow: I caught an error where this toast message could be used, it's when you invite a user who's already in the system. The error should be caught and handled in the back-end and displayed here, but the check was written in the service layer of the API so we should do that :') we can sync on how to fix this

frontend/src/components/common/Toasts.tsx
```diff
@@ -0,0 +1,44 @@
+import { useEffect, useState } from "react";
```
safchow, line 1: you're making reusable components you are a DEMON 👏

safchow (changes requested):
tested the api and it works :D there is an edge case but we should handle it in the back-end but for the most part, this PR is ready 😤 

for the padding and margins on the boxes, I want to see if theres a notable difference in removing them, because our other forms don't use them and it will be inconsistent but we can sync on this

frontend/src/components/forms/CreateEmployee.tsx
```diff
@@ -0,0 +1,320 @@
...
+    setInvitedFirstName(inputValue);
+    if (inputValue) {
+      setInvitedFirstNameError(false);
+    } else {
+      setInvitedFirstNameError(true);
+    }
+  };
+
+  const handleLastNameChange = (e: { target: { value: unknown } }) => {
+    const inputValue: string = e.target.value as string;
```
owen-sellner: This line can be shortened, `const inputValue = e.target.value as string;` also works.

frontend/src/components/forms/CreateEmployee.tsx
```diff
@@ -0,0 +1,320 @@
...
+
+  useEffect(() => {
+    if (invitedAdminStatus === "1") {
+      setIsDisabled(true);
+    } else {
+      setIsDisabled(false);
+    }
+  }, [invitedAdminStatus]);
+
+  const handleFirstNameChange = (e: { target: { value: unknown } }) => {
```
owen-sellner, line 63: Currently users can type numbers into the name input boxes. You can use regex to block them from putting anything other than letters:
```
const handleFirstNameChange = (e: { target: { value: unknown } }) => {
    const inputValue = e.target.value as string;
    if (/^[a-z]{0,}$/i.test(inputValue)) {
      setInvitedFirstName(inputValue);
      setInvitedFirstNameError(false);
    }
  };
```

Also, since there is error validation on submit you don't need `setInvitedFirstNameError(true);`

frontend/src/components/forms/CreateEmployee.tsx
```diff
@@ -0,0 +1,320 @@
...
+    const inputValue: string = e.target.value as string;
+    setInvitedLastName(inputValue);
+    if (inputValue) {
+      setInvitedLastNameError(false);
+    } else {
+      setInvitedLastNameError(true);
+    }
+  };
+
+  const handleEmailChange = (e: { target: { value: unknown } }) => {
```
owen-sellner: This method is never called, is it disabled for testing?

frontend/src/components/forms/CreateEmployee.tsx
```diff
@@ -0,0 +1,320 @@
...
+              >
+                <Text color="#1B2A2C">Require Two Factor Authentication</Text>
+              </Checkbox>
+              <Text fontSize="12px" color="#1B2A2C" fontWeight="bold">
+                Requiring Two Factor Authentication means the employee will only
+                be able to access the platform while physically in the main
+                building
+              </Text>
+            </Box>
+            <Box
```
owen-sellner: Instead of using a box you could simplify it using `<ModalFooter></ModalFooter>` but it would need to be outside of the `ModalBody` tags. This is what it would look like:

image

frontend/src/components/forms/CreateEmployee.tsx
```diff
@@ -0,0 +1,320 @@
...
+              <FormControl isRequired isInvalid={invitedAdminStatusError}>
+                <RadioGroup
+                  value={invitedAdminStatus}
+                  onChange={handleAdminStatusChange}
+                  pb="24px"
+                >
+                  <Flex>
+                    <Box marginRight="24px">
+                      <Radio size="lg" colorScheme="teal" value="1">
+                        <Text color="#1B2A2C">Admin</Text>
```
owen-sellner: I could not figure out what is causing the font difference, if you want a band-aid fix `fontWeight="medium"` makes the text look more like the input text. I don't think the font difference is a huge issue though.

This is what it looks like using `<Text fontWeight="medium" color="#1B2A2C"></Text>`:
image

owen-sellner (changes requested):
I went through the steps to test and everything was working 🥳. I left a couple comments about some styling and input stuff too.

frontend/src/components/common/Toasts.tsx
helenellyx: ### thanks, never expect less from me

frontend/src/components/forms/CreateEmployee.tsx
helenellyx: oops I forgot to remove it

backend/app/rest/residents_routes.py
```diff
@@ -42,7 +42,6 @@ def add_resident():
         return jsonify({"error": (error_message if error_message else str(e))}), 500
 
 
-
```
safchow, line 45: 37 comments

helenellyx: **### `bahahahahahahahhhahaha`**

frontend/src/components/forms/CreateEmployee.tsx
safchow: I fixed the handlers and added them back in as the onchanges

safchow (approved):
LGTM 🐱 thank you for letting me farm contributions 🙏

frontend/src/components/forms/CreateEmployee.tsx
helenellyx: wait @Safewaan lowkey we don't need any of the handlers, bc I am alr checking the error in the `handleSubmit` function. and we can just set up the states directly from the input instead of calling the handlers

frontend/src/components/forms/CreateEmployee.tsx
```diff
@@ -0,0 +1,301 @@
...
+                </RadioGroup>
+              </FormControl>
+              <Checkbox
+                size="lg"
+                colorScheme="teal"
+                borderRadius="40px"
+                isChecked={isTwoFactorAuthenticated}
+                isDisabled={isDisabled}
+                onChange={(event) =>
+                  setIsTwoFactorAuthenticated(event.target.checked)
```
owen-sellner, line 280: I found an edge case where the checkbox remains checked when you switch to Admin from Non Admin. This causes `isTwoFactorAuthenticated` to be true when the form is submitted. 

image

owen-sellner (changes requested):
LGTM! I found one issue but after that is resolved I think it is ready to merge 👏

frontend/src/components/forms/CreateEmployee.tsx
safchow: good catch, I'll add in a fix

owen-sellner (approved):
LGTM!!!!!!

Issue #92, opened 2023-06-05
Edit and Delete Resident Menu


safchow:
## Summary

blocked by designs

## Goals

- [ ] Goal 1
- [ ] Goal 2
- [ ] Goal 3

PR #134, opened 2023-07-08
Log Record Deleting/Editing

owen-sellner:
## GitHub Issue link

[Editing/Deleting Log Records Menu](https://github.com/uwblueprint/supportive-housing/issues/89)

## Implementation description
* Added kebab menu in `LogRecordsTable` for each log record only if role is admin or user is the author
* Created common "DeleteConfirmation` modal for the user to confirm before the item is deleted
* Created `EditLog` from the `CreateLog` form to edit the information in a log record

### Options Menu
1. Verify the kebab menu appears when an admin
2. Verify the kebab menu appears when an author
3. Verify the kebab menu does not appear when not an admin or an author

### Delete Confirmation
5. Verify the delete confirmation modal appears when opened from the menu
6. Verify the log record is only deleted if the user confirms the deletion
7. Verify the correct log record is removed

### Edit Form
8. Verify the edit form appears when opened from the menu
9. Verify the edit form contains the correct details from the log record
10. Check that no changes are made to the log record if the form is not submitted

## What should reviewers focus on?
* Does a `PutLogRecordsResponse` type need to be created in `LogRecordTypes`?

## Checklist
- [x] My PR name is descriptive and in imperative tense
- [x] My commit messages are descriptive and in imperative tense. My commits are atomic and trivial commits are squashed or fixup'd into non-trivial commits
- [x] I have run the appropriate linter(s)
- [x] I have requested a review from the PL, as well as other devs who have background knowledge on this PR or who will be building on top of this PR
- [x] If I have made API changes, I have updated the [REST API Docs](https://www.notion.so/uwblueprintexecs/REST-Endpoints-05ce60312bb943439dfda42bb1318536)
- [x] IF I have made changes to the db/models, I have updated the [Data Models Page](https://www.notion.so/uwblueprintexecs/Data-Models-760f8aa06b244eb0842c079ad77987b0)
- [x] I have updated other Docs as needed

PR #143, opened 2023-07-13
Edit and Delete Resident

danielk1345:
## GitHub Issue link

[Issue Name](https://github.com/uwblueprint/supportive-housing/issues/)
#92 

## Implementation description
* use API to edit and delete employees
* created modal to allow for employee editing and deletion

## Steps to test
1. Add multiple users to the resident database
2. Test editing basic features like initials, move in date, etc
3. Test check boxing and unboxing the resident move out date and seeing if it is able to be edit
4. Test invalid and valid move in/move out date combinations
5. Test deletion

* 

## Checklist


frontend/src/components/pages/HomePage/SearchAndFilters.tsx
danielk1345, line 1: Needed to change the dates to be strings due to the changes to the resident types

frontend/src/components/forms/CreateResident.tsx
```diff
@@ -52,9 +52,10 @@ const CreateResident = (): React.ReactElement => {
 
   const addResident = async () => {
     await ResidentAPIClient.createResident({
+      id: 0,
```
danielk1345: Not sure why but it wouldn't compile unless we passed in an ID (think its because in our resident types, ID is no longer optional). So it is passed as 0 but we don't actually do anything with it so it shouldn't cause any issues

frontend/src/components/pages/ResidentDirectory/ResidentDirectoryTable.tsx
```diff
@@ -58,26 +99,64 @@ const ResidentDirectoryTable = ({
...
+                    <MenuButton
+                      as={IconButton}
+                      aria-label='Options'
+                      icon={<VscKebabVertical />}
+                      w="36px"
+                      variant="ghost"
+                    />
+                    <MenuList>
+                      <MenuItem onClick={() => handleEditToggle(resident.id)}>
+                        Edit Log Record
```
kevin-pierce:
```suggestion
                        Edit Resident
```

frontend/src/components/pages/ResidentDirectory/ResidentDirectoryTable.tsx
```diff
@@ -58,26 +99,64 @@ const ResidentDirectoryTable = ({
...
+                      </MenuItem>
+                      <MenuItem onClick={() => handleDeleteToggle(resident.id)}>
+                        Delete Log Record
```
kevin-pierce:
```suggestion
                        Delete Resident
```

frontend/src/components/forms/EditResident.tsx
```diff
@@ -0,0 +1,314 @@
...
+  const handleBuildingChange = (
+    selectedOption: SingleValue<{ label: string; value: string }>,
+  ) => {
+    if (selectedOption !== null) {
+      setUserBuilding(selectedOption.value);
+      setBuildingError(false);
+    }
+  };
+
+  const handleSubmit = () => {
```
kevin-pierce:
```suggestion
  const handleSave = () => {
```

kevin-pierce: Naming suggestion for clarity

frontend/src/components/forms/EditResident.tsx
```diff
@@ -0,0 +1,314 @@
...
+      initials.length !== 2 ||
+      moveInDateError || 
+      roomNumberError ||
+      userBuilding === "" ||
+      moveOutDateError
+    ) {
+      return;
+    }
+    editRes();
+    setShowAlert(true);
```
kevin-pierce: We need the following to ensure the modal actually closes once we hit save
```suggestion
    setShowAlert(true);
    toggleClose();
```

frontend/src/components/forms/EditResident.tsx
```diff
@@ -0,0 +1,314 @@
...
+                <Col>
+                  <FormControl isRequired isInvalid={initialsError}>
+                    <FormLabel>Resident Initials</FormLabel>
+                    <Input
+                      placeholder={initial}
+                      value={initials}
+                      onChange={handleInitialsChange}
+                    />
+                    <FormErrorMessage>
+                      Resident Initials are required and must contain 2 letters.
```
kevin-pierce, line 198: nit: I'd prefer if we moved the error messages into an enum and simply accessed them using `EditResidentError.RESIDENT_INITIALS` or something along those lines, to help declutter the jsx

frontend/src/components/forms/EditResident.tsx
```diff
@@ -0,0 +1,314 @@
...
+              <Checkbox
+                colorScheme="gray"
+                style={{ paddingTop: "1rem" }}
+                onChange={() => handleFlagged()}
+                marginBottom="16px"
+              >
+                <Text>Edit Move out Date</Text>
+              </Checkbox>
+              <FormControl isRequired isInvalid={moveOutDateError} isDisabled={!flagged}>
+                    <FormLabel>Move out date</FormLabel>
```
kevin-pierce:
```suggestion
                    <FormLabel>Move Out Date</FormLabel>
```

frontend/src/APIClients/ResidentAPIClient.ts
```diff
@@ -67,8 +68,50 @@ const createResident = async ({
   }
 };
 
+const deleteResident = async (residentId: number): Promise<boolean> => {
```
kevin-pierce: nit: We should probably follow the named parameters convention
```suggestion
const deleteResident = async ({residentId}: Pick<Resident, 'residentId'>): Promise<boolean> => {
```

frontend/src/components/common/DeleteConfirmation.tsx
```diff
@@ -0,0 +1,72 @@
...
+          <ModalOverlay />
+          <ModalContent>
+            <ModalHeader>
+              Delete {ITEM_NAME.charAt(0).toUpperCase() + ITEM_NAME.slice(1)}
+            </ModalHeader>
+            <ModalBody>
+              <Box marginBottom="12px">
+                <Text>
+                  Are you sure you want to delete this {ITEM_NAME}? Deleting a{" "}
+                  {ITEM_NAME} will permanently remove it from your system.
```
kevin-pierce:
```suggestion
                  {`Are you sure you want to delete this {ITEM_NAME}? Deleting a 
                  ${ITEM_NAME} will permanently remove it from your system.`}
```

frontend/src/components/forms/EditResident.tsx
kevin-pierce: As a follow up here, I think we need to restructure the component and its props so that we can trigger a refetch of all residents + the residents count when we initiate an edit OR a delete, since right now neither of these are happening (we don't see entries in the table get updated after either action).

Alternatively, a very quick and dirty solution is to force refresh the page (this will circumvent the issue entirely for now from my understanding)

@Safewaan any ideas on what to do? I think if we want to get this in ASAP then the second option wouldn't be a bad idea to use as a bandaid solution for now

frontend/src/components/pages/ResidentDirectory/ResidentDirectoryTable.tsx
```diff
@@ -38,6 +48,37 @@ const ResidentDirectoryTable = ({
...
+  // Delete confirmation state
+  const [deleteOpenMap, setDeleteOpenMap] = useState<{ [key: number]: boolean }>({});
+
+  // Handle delete confirmation toggle
+  const handleDeleteToggle = (residentId: number) => {
+    setDeleteOpenMap((prevDeleteOpenMap) => ({
+      ...prevDeleteOpenMap,
+      [residentId]: !prevDeleteOpenMap[residentId],
+    }));
+  };
```
kevin-pierce: daniel writing the toggle logic:

frontend/src/components/pages/ResidentDirectory/ResidentDirectoryTable.tsx
```diff
@@ -58,26 +99,64 @@ const ResidentDirectoryTable = ({
               <Th>Building</Th>
               <Th>Residency Start Date</Th>
               <Th>Residency End Date</Th>
+              <Th> </Th>
```
kevin-pierce, line 124: nit: Add a comment here to indicate that this is a spacer

frontend/src/types/ResidentTypes.ts
```diff
@@ -1,10 +1,10 @@
 export type Resident = {
-  id?: number;
+  id: number;
   residentId?: string;
   initial: string;
   roomNum: number;
-  dateJoined: Date;
-  dateLeft?: Date;
+  dateJoined: string;
+  dateLeft?: string;
```
kevin-pierce, line 7: I'm just thinking this might be a breaking change elsewhere in the app ; is this 100% required?

kevin-pierce (changes requested):
Looks good overall, just had those minor nit comments + the restructuring of the components to enable refetching of residents + updating the deletion of a resident based on https://uwblueprint.slack.com/archives/C058Z12QYV9/p1689790547661959?thread_ts=1689777903.495979&cid=C058Z12QYV9

frontend/src/components/forms/EditResident.tsx
```diff
@@ -0,0 +1,314 @@
...
+                  </FormControl>
+                </Col>
+              </Row>
+              <Row style={{ marginTop: "16px", marginBottom: "16px" }}>
+                <Col>
+                  <FormControl isRequired isInvalid={buildingError}>
+                    <FormLabel>Building</FormLabel>
+                    <Select
+                      options={BUILDINGS}
+                      placeholder={building}
```
safchow: we should avoid using placeholders for the select and should stick to defaultValue instead. Here's a [SO](https://stackoverflow.com/questions/43495696/how-to-set-a-default-value-in-react-select) post on how to do so

backend/app/rest/residents_routes.py
```diff
@@ -94,7 +93,10 @@ def delete_resident(resident_id):
         )
     except Exception as e:
         error_message = getattr(e, "message", None)
-        return jsonify({"error": (error_message if error_message else str(e))}), 500
+        if ("has existing log records" in str(e)):
+            return jsonify({"existing_log_record_error": "Resident has existing log records"}), 400
+        else:
+            return jsonify({"error": (error_message if error_message else str(e))}), 500
```
kevin-pierce, line 86: nit: I feel like we could use a simpler string to throw the 400 here, but this part isn't necessary ; The main thing I wanted to point out was since this is an internal error (and the error string is effectively acting in the capacity of a variable), we might want to make it all capital, something like `ERROR_EXISTING_RECORD` or something

frontend/src/components/common/DeleteResidentConfirmation.tsx
```diff
@@ -0,0 +1,75 @@
...
+          <ModalContent>
+            <ModalHeader>
+            {`This is a permanent action.
+              Are you sure you want to delete Resident ${RES_ID}`}
+            </ModalHeader>
+            <ModalBody>
+              <Box marginBottom="12px">
+                <Text>
+                  Residents can only be deleted if there are no log records associated with them.
+                </Text>
```
kevin-pierce: nit: Let's throw both of these into variables above (similar to how Owen did it [in his PR](https://github.com/uwblueprint/supportive-housing/pull/134/files#diff-5436ffa96ac1c08e02707cee29e702b374988920782e980738015591d6bd198aR31-R36))

frontend/src/components/forms/EditResident.tsx
```diff
@@ -0,0 +1,315 @@
...
+                <Col>
+                  <FormControl isRequired isInvalid={roomNumberError}>
+                    <FormLabel>Room Number</FormLabel>
+                    <Input
+                      value={roomNumber}
+                      onChange={handleRoomNumberChange}
+                      type="number"
+                    />
+                    <FormErrorMessage>
+                      Room Number is required and must only contain numbers.
```
kevin-pierce, line 211: nit: same comment as above on the error messages, it might make it nicer if we chuck these into variables above

connor-bechthold (commented):
Couple of things:
- Currently there's a couple errors in some files:
  - Need to import `Resident` used here: https://github.com/uwblueprint/supportive-housing/pull/143/files#:~:text=%7D%3A-,Resident,-)%3A%20Promise
  - Need to remove `id` passed in here: https://github.com/uwblueprint/supportive-housing/pull/143/files#:~:text=createResident%20%3D%20async%20(%7B-,id,-%2C
  
- If I try and edit a resident -> check that I want to edit -> and then submit, it lets me submit even though I didn't include anything in the field. This causes weird behaviour when you reopen the modal, such as this:
Screenshot 2023-07-30 at 1 29 55 AM\
- When the modal closes and reopens, the state of the previous options I set are not reset (this should be a function `onClose` that clears these values. Here's a screenshot of an example of this that again causes weird behaviour
Screenshot 2023-07-30 at 1 32 45 AM\
- I think it would be beneficial to separate the error message `Move out date is required and must be after move in date` into two separate messages
  - If you check the box and submit with an empty field, you get `Move out date is required`  
  - If you check the box and submit with a date before the move in date, you get `Move out date must be after move in date`
- It looks like the spacing on the bar here is off (not sure where the Figma design is referenced THERE'S SO MANY PAGES 😭)
Screenshot 2023-07-30 at 1 38 26 AM\ 
- @kevin-pierce mentioned that there's some functionality missing (i.e. modal auto-closing), so I would definitely try and prioritize that before any of the "nitpick" comments. Definitely sync with him on that if you need any clarification (I would help but I'm busy tomorrow 🥲 )
- Take a look at https://github.com/uwblueprint/supportive-housing/pull/134 if you need any guidance/inspiration with anything! The functionality is complete here, and very similar to what this PR is going to accomplish, just with log records instead of residents

connor-bechthold:
So this is looking better, but there's still an issue of the data in the modal matching the row. If you have time to try, occasionally you'll find that when you try and edit, the data populated in the modal won't match that corresponding row, and I believe it has to do with how we're passing data to each modal component.

There was a way to fix this in Owen's PR, but I'm realizing now that there's a way simpler way to do this. We're actually able to instantiate ONE edit and delete modal, and simply pass in the correct data to it when each row is clicked.

To do this, each row button will have a new onClick function. I'll give an example of the edit modal, but the delete modal is the exact same:
`onClick={() => handleEditClick(resident)}`

We'll also have a new state variable `editingResident` which will store the current resident that we want to show in the modal. We'll still have a state variable `isEditModalOpen`. `handleEditClick` will look something like this:

```
const handleEditClick = (resident) => {
  setEditingResident(resident)
  setIsEditModalOpen(true)
}
```
- When a row is clicked, we set the current resident we're editing in state
- We then set the edit modal to open

The edit modal component itself will then take in `editingResident` and `isEditModalOpen` like we currently have, and any other state that is needed. That way we can have one instance each for these modals, and not have to keep track of variables for each modal.

Note that the modal components themselves are not changing, just how they're instantiated. This should fix the above issue, and is also way more efficient to render.

frontend/src/components/forms/CreateResident.tsx
```diff
@@ -50,12 +50,13 @@ const CreateResident = (): React.ReactElement => {
...
   const [showAlert, setShowAlert] = useState(false);
 
+  const ROOM_ERROR_TEXT = `Room Number is required and must only contain numbers.`;
   const addResident = async () => {
     await ResidentAPIClient.createResident({
       initial: initials.toUpperCase(),
       roomNum: parseInt(roomNumber, 10),
-      dateJoined: moveInDate,
-      building,
+      dateJoined: moveInDate.toLocaleDateString('en-CA'),
```
connor-bechthold: This should be `.toISOString()` to ensure it's in UTC

frontend/src/components/forms/EditResident.tsx
```diff
@@ -0,0 +1,345 @@
...
+  const editRes = async () => {
+    await ResidentAPIClient.editResident({
+      id,
+      residentId,
+      initial: initials.toUpperCase(),
+      roomNum: roomNumber,
+      dateJoined: moveInDate.toISOString(),
+      building: userBuilding,
+      dateLeft: moveOutDate ? moveOutDate.toISOString() : undefined,
+    }).then((res) => {
```
connor-bechthold: To keep things consistent through the app can you get rid of the `.then()` callback?
`const res = await ....`
`if (res).....`

frontend/src/components/forms/EditResident.tsx
```diff
@@ -0,0 +1,345 @@
...
+        setAlertData(ALERT_DATA.ERROR);
+      }
+    });
+  };
+
+  const handleClear = () => {
+    setMoveOutDate(undefined);
+  };
+  
+  const handleInitialsChange = (e: { target: { value: unknown } }) => {
```
connor-bechthold: Can use `e: React.ChangeEvent<HTMLTextAreaElement>` instead

frontend/src/components/forms/EditResident.tsx
```diff
@@ -0,0 +1,345 @@
...
+      setInitialsError(false);
+    }
+  };
+
+  const handleFlagged = () => {
+    setFlagged(!flagged);
+    setMoveOutDate(dateLeft? new Date(Date.parse(dateLeft)) : undefined);
+  };
+
+  const handleRoomNumberChange = (e: { target: { value: unknown } }) => {
```
connor-bechthold: Same comment as above

frontend/src/components/forms/EditResident.tsx
```diff
@@ -0,0 +1,345 @@
...
+              <Checkbox
+                colorScheme="gray"
+                style={{ paddingTop: "1rem" }}
+                onChange={() => handleFlagged()}
+                marginBottom="16px"
+              >
+                <Text>Edit Move Out Date</Text>
+              </Checkbox>
+              <FormControl isRequired isInvalid={moveOutDateError} isDisabled={!flagged}>
+                    <FormLabel>Move Out date</FormLabel>
```
connor-bechthold: Can this be `Move Out Date` to keep things consistent?

frontend/src/components/forms/EditResident.tsx
```diff
@@ -0,0 +1,345 @@
...
+                    <SingleDatepicker
+                      name="date-input"
+                      date={moveOutDate}
+                      onDateChange={handleMoveOutDateChange}
+                      propsConfigs={singleDatePickerStyle}
+                    />
+                    <FormErrorMessage marginBottom="8px">
+                      Move out Date is required and must be after Move in Date
```
connor-bechthold: Can this just be ` Move out Date must be after Move in Date`? Remember it's not required, can be left blank and optional

frontend/src/components/forms/EditResident.tsx
```diff
@@ -0,0 +1,345 @@
...
+              </Row>
+              <Checkbox
+                colorScheme="gray"
+                style={{ paddingTop: "1rem" }}
+                onChange={() => handleFlagged()}
+                marginBottom="16px"
+              >
+                <Text>Edit Move Out Date</Text>
+              </Checkbox>
+              <FormControl isRequired isInvalid={moveOutDateError} isDisabled={!flagged}>
```
connor-bechthold: I don't think this should be required. It can be optional, and the red star is a bit misleading here:
Screenshot 2023-08-01 at 11 54 34 PM

connor-bechthold (changes requested):
Couple more bugs I noticed after fixing the timezone error
- The room number validation doesn't work if I submit an empty room number (an error is thrown by the server)
- I can select a move out date that's the same date as the move in date, which shouldn't be allowed (only strictly after)
Screenshot 2023-08-02 at 12 12 01 AM\
- Edit the move out date and make it invalid. If you then uncheck, the error message is still there, even though it should disappear
Screenshot 2023-08-01 at 11 55 44 PM\
- The error message for the move out date is persistent which I can see being annoying. If you have a valid date, and then make it invalid, the error message appears, but the date doesn't actually change. Thus, you have to change the date again to have it disappear. The flow should be:
  - The user can change to any date they want
  - If they press submit and the move out date is invalid, the error message appears
  - As soon as they change the data again, the error message disappears
- Make sure you also handle the case where an error appears because the end date is before the start date, but instead of changing the end date, you change the start date. The error should still disappear. This also should be applied for the vice versa case (error with start date being after end date, but instead of changing the start date, you change the end date)

I also left a couple of comments to look at. This PR is tough because there's a LOT of different edge cases to consider. IT'S ALMOST THERE THOUGH!!!!!!!!!!!!

PR #150, opened 2023-07-18
get residents and employees from db

### LINKED ISSUE #87: Search and FIlters BE/FE Integration
## Summary

blocked by #72 
input components need to be updated to select library used in create log

## Goals

- [ ] Goal 1
- [ ] Goal 2
- [ ] Goal 3

phamkelly17:
## GitHub Issue link
https://github.com/uwblueprint/supportive-housing/issues/87

## Implementation description
* use API to get employees
* have employees from db in the filter dropdown for employees and attention tos
* use API to get residents
* have residents from db in the filter dropdown for residents 

## Steps to test
1. In the filter, check if all the users in the db are included in the select options for Employee
2. Check if the filtering works for Employee
3. Repeat steps 1 + 2 for Attention To 
4. In the filter check is all residents in the db are included in the select option for Resident
5. Check if the filtering works for residents

* 

## Checklist


frontend/src/components/pages/HomePage/SearchAndFilters.tsx
```diff
@@ -103,6 +76,43 @@ const SearchAndFilters = ({
...
+      const labels = users.map((user: User) => {
+        return {
+          id: user.id,
+          label: `${user.firstName} ${user.lastName}`,
+          value: user.id,
+        } as UserLabel;
+      });
+      setUserLabels(labels);
+    } else {
+      setUserLabels(undefined);
```
connor-bechthold: Don't need to include this else clause, this is only called when the component renders and the state is already undefined above on line 79

frontend/src/components/pages/HomePage/SearchAndFilters.tsx
```diff
@@ -103,6 +76,43 @@ const SearchAndFilters = ({
...
+        return {
+          id: resident.id,
+          label: `${resident.initial}${resident.roomNum}`,
+          value: resident.id,
+        } as ResidentLabel;
+      });
+      setResidentLabels(labels);
+    } else {
+      setResidentLabels(undefined);
+    }
```
connor-bechthold, line 107: Same comment as above

frontend/src/components/pages/HomePage/SearchAndFilters.tsx
```diff
@@ -103,6 +76,43 @@ const SearchAndFilters = ({
...
+  };
+
+  const getResidents = async () => {
+    const data = await ResidentAPIClient.getResidents({ returnAll: true });
+    const residentsData = data?.residents;
+    if (residentsData) {
+      const labels = residentsData.map((resident: Resident) => {
+        return {
+          id: resident.id,
+          label: `${resident.initial}${resident.roomNum}`,
```
connor-bechthold: Instead of this string concatenation you should use `resident.residentId` which is constructed in this format for you

frontend/src/types/ResidentTypes.ts
```diff
@@ -1,13 +1,17 @@
...
   roomNum: number;
   dateJoined: Date;
   dateLeft?: Date;
   building: string;
-  label?: string;
-  value?: number;
+};
+
+export type ResidentLabel = {
+  id: number;
```
connor-bechthold: Noticed this in a couple places, is there a reason why a label has an ID if the ID is typically going to be the value?

connor-bechthold (changes requested):
stole some contributions because the frontend wouldn't compile so it should work now

left a couple comments, but looks dope and works as expected 👨‍🍳 👨‍🍳 👨‍🍳

safchow (approved):
everything works correctly :) I also stole some contributions, these are the changes I made:
1. The buildings select wasn't working because residents also have a building, the fix was to use `logs.building` instead of just `building`. I also swapped the value and label in the fake data so the correct value was being sent to the backend
2. I disabled the tags select (I tested it and it works correctly! It sends the selected tag(s) to the backend 👍)  since we won't have that ready for the MVP

frontend/src/components/forms/CreateResident.tsx
```diff
@@ -51,12 +51,12 @@ const CreateResident = (): React.ReactElement => {
...
-      initial: initials.toUpperCase(),
-      roomNum: parseInt(roomNumber, 10),
-      dateJoined: moveInDate,
+    await ResidentAPIClient.createResident(
+      initials.toUpperCase(),
+      parseInt(roomNumber, 10),
+      moveInDate,
       building,
-    });
+    );
```
kevin-pierce: nit: Let's preserve the named parameters here

frontend/src/APIClients/LogRecordAPIClient.ts
```diff
@@ -6,10 +6,22 @@ import {
...
+  residentId?: number[];
+  tags?: string[];
+  flagged?: boolean;
+};
+
+type LogRecordFilters = CountLogRecordFilters & {
+  returnAll?: boolean;
+  pageNumber?: number;
+  resultsPerPage?: number;
+};
```
kevin-pierce: nit: Let's move these types back into a dedicated `types` file

frontend/src/APIClients/ResidentAPIClient.ts
```diff
@@ -45,12 +52,12 @@ const countResidents = async (): Promise<CountResidentsResponse> => {
...
-  roomNum,
-  dateJoined,
-  building,
-}: Resident): Promise<boolean> => {
+const createResident = async (
+  initial: string,
+  roomNum: number,
+  dateJoined: Date,
+  building: string,
+): Promise<boolean> => {
```
kevin-pierce: Let's preserve the named params

frontend/src/types/ResidentTypes.ts
safchow: seems like the same is done for userLabel:
```
export type UserLabel = {
  id: number;
  label: string;
  value: number;
};
```

phamkelly17: not really, took id out

frontend/src/types/UserTypes.ts
```diff
@@ -9,7 +9,6 @@ export type User = {
 };
 
 export type UserLabel = {
-  id: number;
```
kevin-pierce, line 12: I think we'll need to revert this, as above, `getCurUserSelectOption` still returns an object with an `id` field, and feeds this into the state prop which takes in a `UserLabel`, so we'd have misaligned types here

frontend/src/types/LogRecordTypes.ts
```diff
@@ -23,6 +23,27 @@ export type GetLogRecordCountResponse = {
...
-LogRecord, 
-'logId' | 'attnToFirstName' | 'attnToLastName' | 'employeeFirstName' | 'employeeLastName' | 'tags'
+  LogRecord,
+  | "logId"
+  | "attnToFirstName"
+  | "attnToLastName"
+  | "employeeFirstName"
+  | "employeeLastName"
+  | "tags"
 > | null;
```
kevin-pierce: nit: Might be easier / cleaner to convert this to a `pick` type and only include the fields we're concerned with

frontend/src/types/UserTypes.ts
phamkelly17: i just removed the returned id from `getCurUserSelectOption` (since the id in UserLabel isn't doing anything)

kevin-pierce (approved):
LGTM

connor-bechthold (approved):

PR #163, opened 2023-08-01
Split CSV Export date range 

owen-sellner:
## GitHub Issue link

[Issue Name](https://github.com/uwblueprint/supportive-housing/issues/)

There is no issue for this. The goal is to switch the date range input to be like what is used for filtering. 

## Implementation description
* Updated modal in `ExportCSVButton` to use the same date input style as `SearchAndFilters`
* Added error validation to the modal in `ExportCSVButton`  is the start date is before the end date
* Changed `log_records_service` to accept date ranges with only a start date or only an end date 

## Steps to test
1. Create new logs with a variety of dates
2. Verify that CSV export works with no start date and no end date
3. Verify that CSV export works with only start date 
4. Verify that CSV export works with only end date 
5. Verify that CSV export works with both start date and end date
6. Check that the clear button works correctly 

## Images:
image
image

## Checklist

### DIFF
--- a/frontend/src/components/common/ExportCSVButton.tsx
+++ b/frontend/src/components/common/ExportCSVButton.tsx
@@ -11,7 +11,7 @@ import {
   FormControl,
   Button,
   FormHelperText,
-  Flex,
+  FormErrorMessage,
   ScaleFade,
   Alert,
   AlertDescription,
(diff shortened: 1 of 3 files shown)


helenellyx (commented):
everything looks good, I just have a tiny concern. Do we really want the warning to show up if start date and end date are the same? 🤔
Screenshot 2023-08-14 at 9 24 59 PM

owen-sellner:
> 

Oh true, yeah it should not be showing the warning. Thanks for finding that issue.

helenellyx (approved):
yipee looks good

frontend/src/components/common/ExportCSVButton.tsx
```diff
@@ -11,47 +11,66 @@ import {
...
-      date
+  const formatDate = (date: Date | undefined) => {
+    if (date !== undefined) {
+      return date
         .toLocaleString("fr-CA", { timeZone: "America/Toronto" })
-        .substring(0, 10),
-    );
+        .substring(0, 10);
+    }
+    return "";
```
connor-bechthold, line 60: More correct to use `null` here instead of empty quotes since there's no date provided. Don't need to update but if you do, make sure to update `log_records_service` as well

connor-bechthold (approved):
Looks sick! One thing that may be worth adding is a check when there's no records for the selected range. In that case, we're just downloading an empty CSV. Can be made into a follow-up ticket 🔥 🚒

PR #173, opened 2023-09-28
LogRecord Sub-Types

owen-sellner:
## Notion task link

[LogRecord Sub-Types](https://www.notion.so/uwblueprintexecs/LogRecord-Sub-Types-55aeec1f435a4552ada938a697263c5d)

## Implementation description
*  Changed `LogRecord` from a type to an interface
* Create sub-types `AttnToDetails` and `EmployeeDetails` as interfaces and used them to extend the LogRecord interface

## Steps to test
1. Check that records can still be filtered 
2. Check that adding, deleting, and editing records works

* Are there any consequences to changing `LogRecord` to an interface instead of a type?

## Checklist

### DIFF
--- a/frontend/src/types/LogRecordTypes.ts
+++ b/frontend/src/types/LogRecordTypes.ts
@@ -1,17 +1,13 @@
-interface AttnToDetails {
-  attnTo?: number;
-  attnToFirstName?: string;
-  attnToLastName?: string;
+type NameRecord = {
+  id: number;
+  firstName: string;
+  lastName: string;
 };
 
-interface EmployeeDetails {
-  employeeId: number;
-  employeeFirstName: string;
-  employeeLastName: string;
-};
-
-export interface LogRecord extends AttnToDetails, EmployeeDetails {
+export type LogRecord = {
   logId: number;
+  attnTo?: NameRecord;
+  employee: NameRecord;
   building: string;
   datetime: string;
   flagged: boolean;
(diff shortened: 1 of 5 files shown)


braydonwang (approved):
LGTM! Tested filtering and everything still works

I think there may be some linting errors that need to be addressed though (unnecessary semicolon) 🤷

frontend/src/types/LogRecordTypes.ts
```diff
@@ -1,13 +1,19 @@
...
-  datetime: string;
+};
+
+interface EmployeeDetails {
   employeeId: number;
   employeeFirstName: string;
   employeeLastName: string;
+};
+
+export interface LogRecord extends AttnToDetails, EmployeeDetails {
```
connor-bechthold: So instead of extending this, it's better practice to actually just define it as the property type. It would look like this:
```
attnTo: AttnToDetails
employee: EmployeeDetails
```

Then, in the separate interfaces themselves, you can define them like:
```
interface EmployeeDetails {
  id: number;
  firstName: string;
  lastName: string;
}
```

and the same would go for `AttnToDetails`

connor-bechthold (changes requested):
Looks like the linter is hating on your code as well which is failing some checks

frontend/src/types/LogRecordTypes.ts
owen-sellner: Would that make it into a nested object though?

connor-bechthold (approved):
BEAUTIFUL

PR #174, opened 2023-10-01
Delete Tags Route

owen-sellner:
## Notion task link

[Delete Tags Route](https://www.notion.so/uwblueprintexecs/Delete-Tags-Route-e82f96a09c944d2686be56b64b31e652)

## Implementation description
* Create the delete route in `tags_routes.py`
* Added the delete tag function in `tags_service.py` to change the status of a tag from "Active" to "Deleted"

## Steps to test
1. Insert multiple tags in the tags table using `INSERT INTO tags ('name', 'status') VALUE ('example name', 'Active');`
2. Use postman to test the route `http://localhost:8080/tags/delete/` as a put request
Example: `http://localhost:8080/tags/delete/1`
3. Check that the status of the tag has changed in database with `SELECT * FROM tags;` 

## Checklist

### DIFF
--- a/backend/app/rest/tags_routes.py
+++ b/backend/app/rest/tags_routes.py
@@ -18,3 +18,26 @@ def get_tags():
     except Exception as e:
         error_message = getattr(e, "message", None)
         return jsonify({"error": (error_message if error_message else str(e))}), 500
+    
+
+@blueprint.route("/delete/<int:tag_id>", methods=["PUT"], strict_slashes=False)
+@require_authorization_by_role({"Admin"})
+def delete_tag(tag_id):
+    """
+    Delete a tag based on tag id
+    """
+    try:
+        tags_service.delete_tag(tag_id)
+        return (
+            jsonify(
+                {
+                    "message": "Tag with id {tag_id} deleted sucessfully".format(
+                        tag_id=tag_id
+                    )
+                }
+            ),
+            201,
+        )
+    except Exception as e:
+        error_message = getattr(e, "message", None)
+        return jsonify({"error": (error_message if error_message else str(e))}), 500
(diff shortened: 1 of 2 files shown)


braydonwang:
Tested it on postman and everything works as intended! Overall PR LGTM, the only thing I'd say would be to confirm if we want the method type to be "DELETE" instead of "PUT" (and if we do change it to that, we should also change the route name to just "/<<int:tag_id>>"

backend/app/rest/tags_routes.py
```diff
@@ -18,3 +18,26 @@ def get_tags():
     except Exception as e:
         error_message = getattr(e, "message", None)
         return jsonify({"error": (error_message if error_message else str(e))}), 500
+    
+
+@blueprint.route("/delete/<int:tag_id>", methods=["PUT"], strict_slashes=False)
```
connor-bechthold: So you technically did do this right with the PUT (since we're not actually "deleting" anything), but this should be changed into a DELETE route for now. The plan is to remove the `status` field in a follow up PR which I talked about at our last work session, so I think it's more convenient to do it now than later.

Also for the route name, you can keep it consistent with what's been done for other DELETE routes, for example:
```
@blueprint.route("/<int:delete_id>", methods=["DELETE"], strict_slashes=False)
@require_authorization_by_role({"Relief Staff", "Regular Staff", "Admin"})
def delete_log_record(delete_id):
```

connor-bechthold (changes requested):

connor-bechthold (approved):
beauty, LGTM 📦 (just make sure you resolve the merge conflicts before you merge, let me know if you need any help with that!)

phamkelly17 (approved):
nice

PR #175, opened 2023-10-01
Login Page - Functionality

kevin-pierce:
## Notion task link

[Login Page Functionality](https://www.notion.so/uwblueprintexecs/2792c3504d744d2b90ff967c32d00961?v=d33629815ecf477ab4e1ff4e33774c32&p=58fb61c72def4698bbeca6160d73b850&pm=s)

## Implementation description
- Added new exception types to address `TooManyLoginAttempts` and `IncorrectPassword` errors
- Updated exception handling logic on backend to throw Firebase-auth specific exceptions
- Updated AuthClient on FE to throw new `LoginErrorResponse` type when error is encountered
- Updated login form to leverage `FormControl` + related Chakra components

### Incorrect Login Flow:
1. Login to account with incorrect password => Should encounter **Incorrect password** err message
2. Verify that error message is cleared after editing the password (remove / add character)

### Too many Login Attempts Flow:
1. Login to account with incorrect password multiple times => Should encounter **Too many login attempts** err message
2. Verify that error message is cleared after editing the password (remove / add character)

### Invalid Email Flow:
1. Input an invalid email (that doesn't adhere to the specified regex - just omit a `@` symbol for ease) and ensure that the error is continuously displayed _until_ the string is valid
2. Verify that the error is not displayed on page load

### Normal flow
1. Login to account with correct password (should proceed as normal)

Best practices on the propagation of errors from the backend and how they're handled on the frontend (wrt to the displayed error message) is something I'm not too sure about. Would appreciate input into where we could offload more work! (for e.g., we _could_ directly return the error string from the backend and then directly render whatever message is returned from the BE on the FE as opposed to returning an internal err message like "INVALID_PASSWORD")

Error States and Error Messages:

Screen Shot 2023-10-01 at 5 10 31 PM

Screen Shot 2023-10-01 at 5 11 14 PM

Screen Shot 2023-10-01 at 5 11 43 PM

## Checklist


connor-bechthold:
So as a general comment I did some investigating, and I think I found a pretty clean way that we can do error handling a lot better. Flask has this thing called "error decorators" that will automatically send back responses if certain error exceptions are thrown.

Here's a rough example of how I did it:
1. Make a new route file with the error handler:
```
from flask import jsonify

# Error handlers for custom exception types
def handle_custom_error(error):
    return jsonify({"error": "THIS IS A CUSTOM ERROR"}), 400
```

2. Make an exception file with the exception you'd like to define (keeping it pretty standard and simple like you did)
```
class CustomError(Exception):
    def __init__(self, message):
        super().__init__()
        self.message = message
```

3. Register the error handler when we initialize. If you go into `rest/init.py`, you can register the route as an error handler, and specify the exception you want it to trigger under
```
    from .error_handlers import handle_custom_error
    from ..resources.exceptions import CustomError

    app.register_blueprint(user_routes.blueprint)
    app.register_blueprint(auth_routes.blueprint)
    app.register_blueprint(entity_routes.blueprint)
    app.register_blueprint(documentation_routes.blueprint)
    app.register_blueprint(sign_in_logs_routes.blueprint)
    app.register_blueprint(log_records_routes.blueprint)
    app.register_blueprint(residents_routes.blueprint)

    #Exceptions
    app.errorhandler(CustomError)(handle_custom_error)
```

Now whenever in the application we throw some exceptions (whether this be custom or not), we can catch it, and then throw something like a `CustomError` with the specified message. 

I think this can be setup to be pretty useful and powerful for us, especially if we have error handlers/exceptions for specific error codes. I think the big thing here is ensuring that the response to the client is consistent. That way, in axios when we want to parse the error, it's as simple as reading the error code and displaying the error. Ideally, we shouldn't need to do any parsing/error message generation on the frontend.

Let me know what you think! I think in this case, when you catch those custom Firebase exceptions you made, you can then take whatever message is associated with them and throw a `400` or `404` with the technique mentioned above, so the message is easily propagated along

backend/app/rest/auth_routes.py
```diff
@@ -79,6 +83,8 @@ def login():
             **cookie_options,
         )
         return response, 200
+    except (InvalidPasswordException, TooManyLoginAttemptsException) as e:
+        return jsonify({"error": str(e)}), 401
```
connor-bechthold, line 87: Going off my suggestion, we catch and simply `raise 401Error(...whatever message)`

backend/app/rest/auth_routes.py
```diff
@@ -79,6 +83,8 @@ def login():
...
     except Exception as e:
         error_message = getattr(e, "message", None)
```
connor-bechthold, line 89: `raise 500Error(something)`. We could even define a default in the Exception class so we don't need to worry about that here (this is kind of off topic from the point of the PR mb lol)

backend/app/services/implementations/auth_service.py
```diff
@@ -33,6 +37,20 @@ def generate_token(self, email, password):
             token = self.firebase_rest_client.sign_in_with_password(email, password)
             user = self.user_service.get_user_by_email(email)
             return AuthDTO(**{**token.__dict__, **user.__dict__})
+        except InvalidPasswordException as e:
+            self.logger.error(
+                "Incorrect password entered for user with email {email}".format(
```
connor-bechthold, line 42: Adding the logging here is 🤌

backend/app/utilities/exceptions/firebase_exceptions.py
```diff
@@ -0,0 +1,16 @@
+class InvalidPasswordException(Exception):
+    """
+    Raised when an invalid password is entered
+    """
+    def __init__(self, message="INVALID_PASSWORD"):
```
connor-bechthold: So for the message, I would rather this be the actual message we want to display on the frontend, i.e., `"Incorrect password. Please try again."`. What do you think? The main issue I have this is we now have a f/e b/e dependancy where if this code is modified for some reason it won't be detected on the f/e unless it's subsequently changed

backend/app/utilities/firebase_rest_client.py
```diff
@@ -50,7 +55,15 @@ def sign_in_with_password(self, email, password):
 
         response_json = response.json()
 
-        if response.status_code != 200:
+        # Raise an invalid password exception
+        # The corresponding error message from Firebase is INVALID_PASSWORD
+        if ("error" in response_json and response_json["error"]["code"] == 400 and response_json["error"]["message"] == "INVALID_PASSWORD"):
+            raise InvalidPasswordException
+        # The corresponding error message from Firebase is 
+        # 'TOO_MANY_ATTEMPTS_TRY_LATER : Access to this account has been temporarily disabled due to many failed login attempts. You can immediately restore it by resetting your password or you can try again later.'
```
connor-bechthold, line 63: This is an interesting case if this error appears, only because we don't actually provide a user the chance to change their password. Do you know how long the timeout for this is?

frontend/src/APIClients/AuthAPIClient.ts
```diff
@@ -23,20 +25,17 @@ const login = async (
...
-      { withCredentials: true },
-    );
-    return data;
-  } catch (error) {
-    return null;
+    const axiosErr = (error as any) as AxiosError;
+    if (axiosErr.response && axiosErr.response.status === 401) {
+      return {
+        errCode: axiosErr.response.status,
+        errMessage: getLoginErrMessage(axiosErr.response)
```
connor-bechthold, line 32: Going off the above comments, we can just pop the response message here and not have to parse at all

frontend/src/types/AuthTypes.ts
```diff
@@ -5,7 +5,7 @@ export type TwoFaResponse = {
 export type LoginResponse = {
   requiresTwoFa: boolean;
   authUser: AuthenticatedUser;
-} | null;
+} | null | LoginErrorResponse;
```
connor-bechthold: Can we split `LoginResponse` and `LoginErrorResponse` into different types? It would probably be more appropriate to name this to something more general like `ErrorResponse` as well

frontend/src/components/auth/Credentials.tsx
```diff
@@ -40,6 +40,27 @@ const Credentials = ({
 }: CredentialsProps): React.ReactElement => {
   const { authenticatedUser, setAuthenticatedUser } = useContext(AuthContext);
   const history = useHistory();
+  const [emailError, setEmailError] = useState<boolean>(false);
+  const [passwordError, setPasswordError] = useState<boolean>(false);
+  const [passwordErrorStr, setPasswordErrStr] = useState<string>("");
+
+  const handleEmailChange = (e: { target: { value: unknown } }) => {
```
connor-bechthold: nit: you can use `React.ChangeEvent<HTMLTextAreaElement>` instead

frontend/src/components/auth/Credentials.tsx
```diff
@@ -40,6 +40,27 @@ const Credentials = ({
...
+    const emailRegex = /^[a-zA-Z0-9._%+-]+@[a-zA-Z0-9.-]+\.[a-zA-Z]{2,}$/;
+    if (emailRegex.test(inputValue)) {
+      setEmailError(false)
+    } else {
+      setEmailError(true)
+    }
+    setEmail(inputValue)
+  };
+
+  const handlePasswordChange = (e: { target: { value: unknown } }) => {
```
connor-bechthold: same comment as above

frontend/src/components/auth/Credentials.tsx
```diff
@@ -48,7 +69,12 @@ const Credentials = ({
         email,
         password,
       );
-      if (loginResponse) {
+      if (isLoginErrorResponse(loginResponse)) {
+        console.log(loginResponse)
```
connor-bechthold: this can be removed

frontend/src/components/auth/Credentials.tsx
```diff
@@ -97,20 +107,27 @@ const Credentials = ({
...
-              placeholder="username@domain.com"
-            />
+            <FormControl isRequired isInvalid={emailError}>
+              <Input 
+                type="email" 
+                value={email} 
+                onChange={handleEmailChange} 
+                placeholder="Your email address"
+              /> 
+              <FormErrorMessage>Please enter a valid email.</FormErrorMessage>
```
connor-bechthold: lovely stuff right here

connor-bechthold (changes requested):
This is really well done! Left a couple comments with regards to error handling but everything works correctly. 

I noticed if I don't provide any email at all I get this popup: 
Screenshot 2023-10-09 at 2 19 59 AM

Also, with the email validation (I had to do this at my last co-op sadly), it's better if we only display an error after the user has clicked login at least once. It's a bit annoying if they immediately get this error right off the bat, since you have to type an invalid email to get to a valid one.

Would we be able to display an error if the user isn't invited (instead of the popup), or would this require more investigation? I notice when I enter some invalid email I get a generic `Error logging in. Please try again later.` message.

backend/app/utilities/firebase_rest_client.py
kevin-pierce: Unfortunately there's no official documentation on the timeout for this is (not in the Firebase docs) ; all I know is that after some time, Firebase will automatically "undisable" an account, and so the error will stop being thrown

frontend/src/types/AuthTypes.ts
kevin-pierce: For sure ; will update 👀

backend/app/utilities/exceptions/firebase_exceptions.py
kevin-pierce: Yeah that makes sense ; there's definitely more safety with simply just propagating the error message to the FE ; will update

kevin-pierce: Updated

frontend/src/APIClients/AuthAPIClient.ts
kevin-pierce: Updated

backend/app/rest/auth_routes.py
kevin-pierce: I think for now this is fine ; ATM we'd simply display a "Error logging in. Please try again later." message when anything that's not a 401 is returned (and we only return a 401 in the above 2 cases)

frontend/src/components/auth/Credentials.tsx
```diff
@@ -104,43 +110,37 @@ const Credentials = ({
...
-            w="36%"
-            top="27%"
-            left="6%"
-            direction="column"
-            position="absolute"
-            justifyContent="space-between"
-          >
-            <Box display="flex"  alignItems="flex-start">
-              <Text variant="login" position="absolute">
+          <Flex marginTop="270px" display="flex" align="center" justify="center">
```
kevin-pierce: Setting this marginTop isn't ideal, BUT it's required since if we were to wrap this in a flex box that would center itself, then we'd end up in situation where if the error messages are displayed, then the ENTIRE form would get re-centered (and hence get shifted "up" to account for this new height). Using the margin-top ensures that the form is "fixed", while only the input fields + buttons below the error message are shifted

frontend/src/components/auth/Credentials.tsx
```diff
@@ -36,6 +32,10 @@ type CredentialsProps = {
   setToggle: (toggle: boolean) => void;
 };
 
+const isLoginErrorResponse = (res: LoginResponse | ErrorResponse) : res is ErrorResponse => {
```
connor-bechthold: NIFTY

connor-bechthold (changes requested):
This works 💯 💯 💯 💯 💯 ,  a couple small things I noticed could be improved with error display:
- If you get an `Error logging in. Please try again later.` message it only clears if you change the password and not the email. There could be a chance that the email is only slightly off so if you change it the error still appears. Ideally it should also clear the message
- The `Please enter a valid email error` appears as soon as you type, which is a little annoying since when you type your email out it will be invalid until it's valid. What we can do is:
  - Keep some sort of `true/false` state for when the login button is pressed
  - The email error should only appear if there's an invalid email AND the login button has been pressed at least once
  - That way when the user is initially typing their email for the first time, they aren't met with that error
- What would your thoughts be on changing the general error message to `Error logging in. Please try again later.`? I'm only worried that "later" implies they're locked out for a period of time (when they're really not)

carolynzhang18 (approved):
Looks good!! 😎

connor-bechthold (approved):
BEAUTIFUL I think the margin top you added will be fixed with https://github.com/uwblueprint/supportive-housing/pull/184 so this is good to merge

PR #170, opened 2023-09-28
Carolyn/create log sign in

carolynzhang18:
## Notion task link
https://www.notion.so/uwblueprintexecs/Create-Log-On-Sign-In-0364fbd4600f43628cc879bd84b196d2

Create Log On Sign-In (https://www.notion.so/uwblueprintexecs/2792c3504d744d2b90ff967c32d00961?v=d33629815ecf477ab4e1ff4e33774c32)

## Implementation description
* Call create_sign_in_log upon log in (auth_routes.py)
* Add env variable CREATE_SIGN_IN_LOG to control sign in logging

## Steps to test
1. Change env variable CREATE_SIGN_IN_LOG=True.
2. Log in.
3. Check sign_in_log table for new sign-in record.

## Checklist
- [x] My PR name is descriptive and in imperative tense
- [x] My commit messages are descriptive and in imperative tense. My commits are atomic and trivial commits are squashed or fixup'd into non-trivial commits
- [x] I have run the appropriate linter(s)
- [x] I have requested a review from the PL, as well as other devs who have background knowledge on this PR or who will be building on top of this PR
- [x] If I have made API changes, I have updated the [REST API Docs](https://www.notion.so/uwblueprintexecs/REST-Endpoints-05ce60312bb943439dfda42bb1318536)
- [x] IF I have made changes to the db/models, I have updated the [Data Models Page](https://www.notion.so/uwblueprintexecs/Data-Models-760f8aa06b244eb0842c079ad77987b0)
- [x] I have updated other Docs as needed

PR #180, opened 2023-10-12
Require 2FA On Initial Signup for Relief Staff

carolynzhang18:
## Notion task link

2FA On Initial Signup (https://www.notion.so/uwblueprintexecs/2FA-On-Initial-Signup-f681347ebac54fb0b8bb6fa75c5a34f7)

## Implementation description
* Backend: modify /register route to also return whether the user needs 2FA or not (i.e. Relief Staff or not)
* Frontend: toggle from Signup page to 2FA page upon signup for Relief Staff; redirect to Home page for other users

## Steps to test
1. Login as an Admin and invite 3 new users: (a) an Admin, (b) a Relief staff who requires 2FA (check off this box when creating this user), and (c) a Relief Staff who does not require 2FA (don't check off the box).
2. From the Login page, click "Sign Up" and sign up as each of these 3 users. 
3. For (a) and (c), after signing up, you should be redirected to the Home page immediately. For (b), you should be prompted to complete 2FA with Authy.

## Checklist
- [X] My PR name is descriptive and in imperative tense
- [X] My commit messages are descriptive and in imperative tense. My commits are atomic and trivial commits are squashed or fixup'd into non-trivial commits
- [X] I have run the appropriate linter(s)
- [X] I have requested a review from the PL, as well as other devs who have background knowledge on this PR or who will be building on top of this PR
- [X] If I have made API changes, I have updated the [REST API Docs](https://www.notion.so/uwblueprintexecs/REST-Endpoints-05ce60312bb943439dfda42bb1318536)
- [X] IF I have made changes to the db/models, I have updated the [Data Models Page](https://www.notion.so/uwblueprintexecs/Data-Models-760f8aa06b244eb0842c079ad77987b0)
- [X] I have documented this PR in the [Production Release Page](https://www.notion.so/uwblueprintexecs/fe28a97bfa5648d3bc3795b32b694acd?v=5565cfb35dcc45bb84f6528cb09517cd)
- [X] I have updated other Docs as needed


backend/app/rest/auth_routes.py
```diff
@@ -70,6 +70,8 @@ def login():
             "role": auth_dto.role,
         }
 
+        sign_in_logs_service.create_sign_in_log(auth_dto.id)
```
connor-bechthold, line 73: I think there's some sign-in logs changes here in the diff that are from this PR (https://github.com/uwblueprint/supportive-housing/pull/170)

Should be fixed by `git pull main` (from `main`), `git checkout carolyn/2fa-on-initial-signup` and then `git merge main` (there may be some conflicts to resolve)

connor-bechthold: I think there's some sign-in logs changes here in the diff that are from this PR (https://github.com/uwblueprint/supportive-housing/pull/170)

Should be fixed by `git pull main` (from `main`), `git checkout carolyn/2fa-on-initial-signup` and then `git merge main` (there may be some conflicts to resolve)

frontend/src/components/auth/Signup.tsx
```diff
@@ -17,13 +20,24 @@ const Signup = (): React.ReactElement => {
   const onSignupClick = async () => {
     const isInvited = await commonApiClient.isUserInvited(email);
     if (isInvited) {
-      const user: AuthenticatedUser | null = await authAPIClient.register(
+      const registerResponse: RegisterResponse = await authAPIClient.register(
```
connor-bechthold: The type doesn't need to be included here since it's inferred from the return type of `register` - this can just be:
```const registerResponse =  await authAPIClient.register(...```

File: frontend/src/types/AuthTypes.ts [hunk 3]
```diff
@@ -7,12 +9,17 @@ export type LoginResponse = {
   authUser: AuthenticatedUser;
 } | null;
 
+export type RegisterResponse = {
```
connor-bechthold: There's already a type `LoginResponse` that's the same as this - I'm thinking that can be renamed to something more general like `AuthTokenResponse` and used instead, then we don't need to define a new duplicate type

frontend/src/APIClients/AuthAPIClient.ts
```diff
@@ -96,7 +96,7 @@ const register = async (
   lastName: string,
   email: string,
   password: string,
-): Promise<AuthenticatedUser | null> => {
+): Promise<RegisterResponse> => {
   try {
     const { data } = await baseAPIClient.post(
```
connor-bechthold, line 101: Later on in this file the local storage token is set (even when it's not supposed to be) - basically this allows the user to bypass the 2FA page. You can verify this if you signup, go to the 2FA page, and refresh. All you need to do is remove line 106 and you're chilling

connor-bechthold (changes requested):
Just a couple things to address but otherwise looks great!

One thing I think that would also be beneficial is to update the naming/locations of some files:
- Right now `Login.tsx` holds `Authy.tsx` and `Credentials.tsx`. I think we can rename `Credentials.tsx` to `Login.tsx` and move it under `forms`. Then, we can rename the old `Login.tsx` to `LoginPage.tsx` and move it under `pages/auth`
- The same thing can be done with what you did here, with `Signup.tsx` and `SignupPage.tsx`

frontend/src/components/forms/Signup.tsx
```diff
@@ -0,0 +1,122 @@
...
+  setEmail: (email: string) => void;
+  firstName: string;
+  setFirstName: (firstName: string) => void;
+  lastName: string;
+  setLastName: (lastName: string) => void;
+  password: string;
+  setPassword: (password: string) => void;
+  toggle: boolean;
+  setToggle: (toggle: boolean) => void;
+};
```
phamkelly17, line 21: nit: move this to the `types` folder

connor-bechthold (approved):
LGTM!

One thing that I noticed is we still have the `token` field which is in the `Authy` component because of the sign in with google functionality. Since we're not using that anymore, that can be removed (but we can do that in a later ticket so it's not a blocker)

phamkelly17 (approved):
tested it and everything looks good

PR #184, opened 2023-10-20
Overflow gets cancelled on login | BUG

achandrabalan:
## Implementation description
* I've had this weird bug when I logout and log back in that basically freezes the screen and blocks scrolling on the actual page and only allows it on the table. I've been able to get past this with refreshing the page but that is not very intuitive for the user.

## Steps to test
1. See if you can reproduce the bug on your end, press log out and then sign back in. Does the scroll lock?
2. If so, does this fix it for you?

* I think this fixes it on my side not sure if it's a universal fix tho
* Also are there any risks to setting this here? I'm not too sure how it affects the rest of the app.

## Checklist


frontend/src/components/auth/Credentials.tsx
```diff
@@ -98,9 +98,7 @@ const Credentials = ({
     return <Redirect to={HOME_PAGE} />;
   }
 
-  if (toggle) {
-    // Lock scroll
-    document.body.style.overflow = "hidden"
+  if (toggle) {   
     return (
       <Flex h="100vh">
         <Box w="47%">
```
owen-sellner: The total width of the items in this Box tag is greater than 100% so it is causing overflow. Changing line 161 to `width="64%` fixed the issue for me.

owen-sellner (changes requested):
Great job fixing the bug! I found an issue with horizontal overflow but the solution should be quick to implement.

frontend/src/components/auth/Signup.tsx
```diff
@@ -47,8 +47,6 @@ const Signup = (): React.ReactElement => {
   if (authenticatedUser) {
     return <Redirect to={HOME_PAGE} />;
   }
-  // Lock scroll
-  document.body.style.overflow = "hidden"
   return (
     <Flex h="100vh">
         <Box w="47%">
```
owen-sellner: The vertical and horizontal overflow is happening on this page as well. The same solution of changing the widths so the total is less than 100% works.

owen-sellner (changes requested):
I noticed the same issue as before in `Signup.tsx` but after that I think this is ready.

frontend/src/components/forms/Signup.tsx
```diff
@@ -72,72 +72,59 @@ const Signup = ({
...
-            top="28%"
-            left="6%"
+            h="70%"
+            marginTop="15%"
             direction="column"
-            position="absolute"
+            justify="center"
             justifyContent="space-between"
+            align="center"
           >
```
connor-bechthold, line 156: so we don't actually need to use `marginTop` here if flexbox is utilized correctly - `justify` is actually `justifyContent` and `align` is actually `alignItems`

can do something like this which is a bit more simple
```suggestion
          <Flex
            h="100%"
            direction="column"
            justifyContent="center"
            alignItems="center"
            gap="28px"
          >
```

frontend/src/components/forms/Signup.tsx
```diff
@@ -72,72 +72,59 @@ const Signup = ({
...
-            <Box>
+            <Box mr="58%">
```
connor-bechthold: To get the left align you want you can just use `textAlign` instead:

```suggestion
            <Box w="80%" textAlign="left">
```

frontend/src/components/forms/Signup.tsx
```diff
@@ -158,21 +145,20 @@ const Signup = ({
...
             alignContent="center"
-          >
-            <Text variant="loginSecondary" paddingRight="1.1%">
-              Already have an account?
-            </Text>
-            <Text variant="loginTertiary" onClick={onLogInClick}>
-              Log In Now
-            </Text>
+            marginTop="3%"
+            >
```
connor-bechthold: So this can also be simplified to just use flexbox and not use `margin` and `padding`:
```suggestion
     <Box w="80%">
       <Flex gap="10px">
         <Text variant="loginSecondary">
            Already have an account?
         </Text>
         <Text variant="loginTertiary" onClick={onLogInClick}>
           Log In Now
         </Text>
       </Flex>
     </Box>
```

connor-bechthold (changes requested):
So I noticed that the messages below the buttons for both the Signup and Login page are different - I suggested changes on the Signup Page that fix this that utilize Flexbox and removes the use of margin and padding (which could produce unexpected results on different screen sizes)

To keep things consistent the same changes can also be made on the Login page and then both forms should look identical styling wise and this should be good to go! Let me know if you have any questions/confusions with the suggestions at all.

connor-bechthold (approved):
LOOKS GREAT I left like a million comments on this one so thanks for addressing everything 🙏 

This should be good to merge once https://github.com/uwblueprint/supportive-housing/pull/202 is merged (just because it adds error messages on this page that will cause conflicts, but should be easier to address in this PR)

owen-sellner (approved):
LGTM!!!!

PR #186, opened 2023-10-24
Handling Record Edits on the Employee Directory Page

achandrabalan:
## Implementation description
* Deactivating / Activating employees is now shown immediately, not after refresh
* Adding employees instantly triggers toast, increments value and shows on modal, not after refresh
* Deleting employees instantly triggers toast, decrements value and shows on modal, not after refresh

## Steps to test
Deactivate / Activate:
1. If you have an active employee, click 'deactivate' and 'activate'. You should see button grey out OR turn back to green instantly, not after a refresh

Adding employee:
1. Toast message
2. Employee count increments by one instantly
3. Employee is seen on modal instantly
4. If employee count > numRecordsPerPage, the employee is added to the last page, but the user stays on the first page (As seen with record logs)

Delete Employee:
1. Toast message
2. Employee count decrements by one instantly
3. If you delete the last employee on the page, you are sent to the previous page

* 

## Checklist


owen-sellner (approved):
LGTM!!! It runs pretty slowly for me but that might be an issue with my laptop tbh

achandrabalan:
Okay, found a few major bugs but got them resolved. 
My only concerns are 
1. There's not much consistency between edits on (Log Records, Residents, Employees) as the format is a bit different for each.
2. Calling methods on a toggle close may be finicky? Seems good on my testing but I can see vulnerabilities here. Are there better alternatives to call getRecords() (so I can refresh the table)
3. The order of the logs keeps changing? Refreshing, editing and adding residents sometimes changes the order of employees. Also, adding employees sometimes adds to the top / bottom / random. Is this a bug or expected behaviour?

frontend/src/components/forms/CreateEmployee.tsx
```diff
@@ -25,9 +25,20 @@ import commonApiClient from "../../APIClients/CommonAPIClient";
...
+  userPageNum: number;
+  getRecords: (pageNumber: number) => Promise<void>;
+  setUserPageNum: React.Dispatch<React.SetStateAction<number>>;
+  countUsers: () => Promise<void>;
+}
 const RoleOptions = ["Relief Staff", "Admin", "Regular Staff"];
 
-const CreateEmployee = (): React.ReactElement => {
+const CreateEmployee = ({
+  userPageNum,
```
connor-bechthold: you can remove this since it isn't used anywhere

frontend/src/components/forms/CreateEmployee.tsx
```diff
@@ -109,6 +120,9 @@ const CreateEmployee = (): React.ReactElement => {
 
   const handleClose = () => {
     setIsOpen(false);
+    getRecords(1);
+    setUserPageNum(1);
+    countUsers();
```
connor-bechthold: So we don't want this in the `handleClose` function because it will trigger a refetch of the records even when you close from the modal. We only want this to happen when a successful add happens, so this should be right before line 167 in this file, which is where `handleClose` is called after a successful creation

connor-bechthold:
> Okay, found a few major bugs but got them resolved. My only concerns are
> 
> 1. There's not much consistency between edits on (Log Records, Residents, Employees) as the format is a bit different for each.
> 2. Calling methods on a toggle close may be finicky? Seems good on my testing but I can see vulnerabilities here. Are there better alternatives to call getRecords() (so I can refresh the table)
> 3. The order of the logs keeps changing? Refreshing, editing and adding residents sometimes changes the order of employees. Also, adding employees sometimes adds to the top / bottom / random. Is this a bug or expected behaviour?

So:
1. You probably noticed Log Records is setup different than Residents and Employees. We did that table first so we learned things and applied them to Residents and Employees. Ideally I'd want to update it as well but it's not super high priority atm
2. I left a comment w regards to this on my review
3. We talked about this last work session but we need to sort by last modified which can be addressed in a follow up ticket

connor-bechthold (changes requested):
Beautyyyy left a few comments but otherwise looks good! Great work on getting all of these updated, certainly isn't the most straightforward task

frontend/src/components/pages/AdminControls/EmployeeDirectoryTable.tsx
```diff
@@ -246,7 +264,11 @@ const EmployeeDirectoryTable = ({
           <EditEmployee
             employee={editingEmployee}
             isOpen={isEditModalOpen}
-            toggleClose={() => setIsEditModalOpen(false)}
+            toggleClose={() => {
+              setIsEditModalOpen(false)
+              getRecords(userPageNum)
```
connor-bechthold: So you should be passing this in as a prop to the component because we don't want this to trigger unless the form is submitted (not just closed). This can be added to `EditEmployee` on line 160 before `toggleClose` is called

connor-bechthold (approved):
LGTMMMMM 🚒

PR #194, opened 2023-11-05
Verification Page

owen-sellner:
## Notion task link

[Login/Signup Verification](https://www.notion.so/uwblueprintexecs/Login-Signup-Verification-43263134e4e14481a0de27c0c011a780)

## Implementation description
* Updated `AuthenticatedUser` type to include if the user is verified in `AuthTypes.tsx`
* Updated login and register in `auth_routes.py`
* Created `is_authorized_by_token` in `auth_service.py` to check if a user is verified using the access token
* Created `isVerified` in `AuthAPIClient.ts` for the verification page button to update the AuthenticatedUser context if the user becomes verified

## Steps to test
1. Access the database using `docker exec -it SHOW-database /bin/bash -c "psql -U postgres -d show_db"`
2. Invite a new account using `insert into users (first_name, last_name, role, user_status, email) values ('your first name', 'your last name', 'Admin', 'Invited', 'your email');`
3. Check if after the Signup page the user is taken to the verification page 
4. Check that pressing the button on the Verification page prompts the user with an error toast message
5. Verify account from email
6. Check that pressing the button on the Verification page takes the user to the homepage
7. Logout from the account
8. Sign in to account from Login page and check that the user is taken directly to the homepage 

* Should there be a button for the user to logout from the Verification page to prevent getting stuck if the verification email expires? 

## Screenshot
image

## Checklist


frontend/src/components/auth/PrivateRoute.tsx
```diff
@@ -17,11 +16,23 @@ const PrivateRoute: React.FC<PrivateRouteProps> = ({
...
 
-  return authenticatedUser ? (
+  if (authenticatedUser === null) {
+    return (
+      <Redirect to={LOGIN_PAGE} />
+    )
+  }
+
+  if (authenticatedUser.verified === false) {
+    if (!window.location.pathname.endsWith("/verification")) {
```
connor-bechthold: Nit: React router has a built in way to do this

```
import { useLocation } from 'react-router-dom';
const location = useLocation();
const currentPath = location.pathname;
```

Also I'm assuming you're checking this because otherwise you'll be stuck on an infinite loop on the verification page?

frontend/src/components/auth/Verification.tsx
```diff
@@ -0,0 +1,63 @@
...
+  const history = useHistory();
+  const { authenticatedUser, setAuthenticatedUser } = useContext(AuthContext);
+
+  const handleVerification = async () => {
+    if (authenticatedUser) {
+      const authUser = authenticatedUser;
+      console.log(authUser.verified);
+      authUser.verified = await authAPIClient.isVerified();
+      setAuthenticatedUser(authUser);
+      console.log(authenticatedUser);
```
connor-bechthold: can be removed

frontend/src/components/auth/Verification.tsx
```diff
@@ -0,0 +1,63 @@
...
+
+const Verification = (): React.ReactElement => {
+  const newToast = CreateToast();
+  const history = useHistory();
+  const { authenticatedUser, setAuthenticatedUser } = useContext(AuthContext);
+
+  const handleVerification = async () => {
+    if (authenticatedUser) {
+      const authUser = authenticatedUser;
+      console.log(authUser.verified);
```
connor-bechthold: can be removed

frontend/src/components/forms/Signup.tsx
```diff
@@ -132,6 +133,7 @@ const Signup = ({
     }
   };
 
+
```
connor-bechthold: can remove this newline

frontend/src/components/forms/Signup.tsx
```diff
@@ -140,7 +142,10 @@ const Signup = ({
   };
 
   if (authenticatedUser) {
-    return <Redirect to={HOME_PAGE} />;
+    if (isVerified) {
```
connor-bechthold: Is there a point of `isVerified`? It looks like it's always going to be false regardless

connor-bechthold (approved):
you did a really good job on this future pl moment fr

had a couple nits/housekeeping comments to look at but otherwise this is good to go 👨‍🍳

backend/app/services/implementations/auth_service.py
```diff
@@ -219,3 +219,13 @@ def is_authorized_by_email(self, access_token, requested_email):
...
+        
+    def is_authorized_by_token(self, access_token):
+        try:
+            decoded_id_token = firebase_admin.auth.verify_id_token(
+                access_token, check_revoked=True
+            )
+            firebase_user = firebase_admin.auth.get_user(decoded_id_token["uid"])
+            return firebase_user.email_verified
+        except:
+            return False
```
kevin-pierce: So one q here ; this function returns a `True` / `False`, but when there's an exception it by default returns false. This, however, will not cause the exception to be caught in the calling sites (for instance, in the route calling `is_verified`, if an exception is raised within `is_authorized_by_token`, we will never see it propagate up, and thus will return a 200 still). Just wanted to know if this was intentional ; otherwise, I think we should introduce some custom exceptions to account for the error down here, and throw that instead of returning `False` by default

frontend/src/components/auth/PrivateRoute.tsx
```diff
@@ -16,12 +15,26 @@ const PrivateRoute: React.FC<PrivateRouteProps> = ({
   path,
 }: PrivateRouteProps) => {
   const { authenticatedUser } = useContext(AuthContext);
+  const location = useLocation();
+  const currentPath = location.pathname;
+
+  if (authenticatedUser === null) {
```
kevin-pierce: nit: I think it might be better to check for _all_ possible falsey values here by simply checking for `if (!authenticatedUser)` for extra safety

kevin-pierce (approved):
One minor nit, and a more general q about the exception handling for the verification endpoint ; personally I think that the exception stuff would require a bit of an overhaul, and so perhaps it could be broken out into its own ticket

phamkelly17 (approved):
lgtm

frontend/src/components/auth/PrivateRoute.tsx
connor-bechthold: +1

backend/app/services/implementations/auth_service.py
connor-bechthold: I think we should at least console out an error here (as a temp soln) just so we have some visibility around this in our logs

PR #196, opened 2023-11-06
Get Buildings Route

phamkelly17:
## Notion task link

[Get Buildings Route](https://www.notion.so/uwblueprintexecs/Get-Buildings-Route-e5b2d8b0d0004b2e8754d133fa4ed3a5?pvs=4)

## Implementation description
* added GET endpoint to get all buildings from DB
* replaced hardcoded `BUILDINGS` variable with the buildings fetched from the database

## Steps to test
1. Check if you're able to see the building options on log filter, add log, edit log, add resident, edit resident
2. Check if the functionality works for the places mentioned above

## Checklist


frontend/src/APIClients/BuildingAPIClient.ts
```diff
@@ -0,0 +1,28 @@
+import axios, { AxiosError } from "axios";
+import AUTHENTICATED_USER_KEY from "../constants/AuthConstants";
+import { getLocalStorageObjProperty } from "../utils/LocalStorageUtils";
+import baseAPIClient from "./BaseAPIClient";
+import { GetBuildingsResponse } from "../types/BuildingTypes";
+
+const getBuildings = async (): Promise<GetBuildingsResponse> => {
+  console.log("here");
```
owen-sellner: Unnecessary console log 😱

frontend/src/APIClients/BuildingAPIClient.ts
```diff
@@ -0,0 +1,28 @@
...
+  try {
+    const bearerToken = `Bearer ${getLocalStorageObjProperty(
+      AUTHENTICATED_USER_KEY,
+      "accessToken",
+    )}`;
+    const { data } = await baseAPIClient.get<GetBuildingsResponse>(
+      `/buildings/`,
```
owen-sellner: The second slash is not required, "/buildings" also works

frontend/src/components/forms/CreateLog.tsx
```diff
@@ -207,6 +205,15 @@ const CreateLog = ({ getRecords, countRecords, setUserPageNum }: Props) => {
 
   // fetch resident + employee data for log creation
   const getLogEntryOptions = async () => {
+    const buildingsData = await BuildingAPIClient.getBuildings();
+
+    if (buildingsData && buildingsData.buildings.length !== 0) {
+      const buildingLabels: BuildingLabel[] = buildingsData.buildings.map(
+        (building) => ({ label: building.name!, value: building.id! }),
```
owen-sellner, line 215: Does the `address` property for the building ever get used?

owen-sellner (changes requested):
Very cool 😎 I found a couple nits but the main functionality works great! 

Also, I noticed the notion ticket for this talks about being able to get a specific building using the id, is that no longer part of this ticket?

frontend/src/components/forms/CreateResident.tsx
```diff
@@ -106,9 +102,22 @@ const CreateResident = ({
     }
   };
 
+  const getBuildingsOptions = async () => {
+    const buildingsData = await BuildingAPIClient.getBuildings();
+
+    if (buildingsData && buildingsData.buildings.length !== 0) {
+      const buildingLabels: BuildingLabel[] = buildingsData.buildings.map(
+        (building) => ({ label: building.name!, value: building.id! }),
```
owen-sellner, line 110: I removed the exclamation marks after the properties and it still works. What do the exclamation marks do?

phamkelly17: it asserts that the name or id are not null (ngl i copied it from where else in the code where they did the same thing)

frontend/src/components/forms/CreateLog.tsx
phamkelly17: no, but i figured it was better to not have the backend return a list of `BuildingLabel` objects. My reasoning was that if we ever wanted to access the address for whatever reason, we wouldn't have to make a new endpoint in the backend. I also figured that we would want to use the address eventually because why else would it be there

connor-bechthold (changes requested):
Everything works as expected besides the edit - if you make a new resident/log record and then try and edit it, the building doesn't appear until you close the modal and open it a second time. I think I ran into this issue last term and it had to do with how I was initially setting the state when the component rendered. Should be good to after this is fixed though 👨‍🍳 
Screenshot 2023-11-12 at 12 08 59 AM

frontend/src/components/forms/EditLog.tsx
```diff
@@ -202,6 +209,7 @@ const EditLog = ({
 
   const initializeValues = () => {
     // set state variables
+    getBuildingsOptions();
```
connor-bechthold: so instead of calling this here, you should call this in `LogRecordsTable.tsx`. the issue is that the state doesn't update in time, so when the building ID from the log record is being matched with these options, they don't exist yet (when you close and re-open the options have loaded ands appear)

there's a function there called `getLogEntryOptions` that you can pop this in and then pass it as a prop to this component which should resolve things

connor-bechthold (changes requested):
so with the log record edit, the previous bug I mentioned is still there. i left a comment on how to resolve it and then this should be good to merge

connor-bechthold (approved):
muy bien

owen-sellner (approved):
très bien

PR #198, opened 2023-11-09
Employee Directory - Sort By Last Modified

achandrabalan:
## Notion task link

[Employee Default Sort By Last Modified](https://www.notion.so/uwblueprintexecs/Employee-Default-Sort-By-Last-Modified-f2289846799b40589a9d69d9033d1089)

## Implementation description
* Currently, the employee directory is sorted in random order. With these changes it should sort by last modification date, with the newest / most recently edited logs at the top.

## Steps to test
1. Add employee, they should append to the top
2. Edit employee, they should move to the top
3. Activate / Deactivate employee, they should move to the top

* Now that editing a log brings it to the first page, shouldnt the user also be brought to the first page?

## Checklist


backend/app/models/user.py
```diff
@@ -19,6 +20,7 @@ class User(db.Model):
         nullable=False,
     )
     email = db.Column(db.String, nullable=False)
+    last_modified = db.Column(db.DateTime, default=func.now(), onupdate=func.now())
```
connor-bechthold: So I think the issue with what you were dealing with was that by default when a User was queried, the `last_modified` column was returned. In order to handle this, you had to update the DTOs for it to work.

I did some research and we can actually get around this by using `defer`. What `defer` does is it doesn't return a column from a query by default unless explicitly stated. So, if we set `defer=True` on `last_modified`, we should never get it back from a query to the Users table. We should also set it as non-nullable. So, the model will look something like this:

```suggestion
    last_modified = db.Column(db.DateTime, default=func.now(), onupdate=func.now(), defer=True, nullable=False)
```

Once you do this, you should be able to remove all of the updates to `auth_routes` and `resources` that you made

connor-bechthold (changes requested):
Left a comment on how you can avoid the issues you were dealing with without having to update all the DTO's - let me know if it makes sense! Once this is done this should be good to merge 👨‍🍳

backend/app/models/user.py
```diff
@@ -1,5 +1,6 @@
 from sqlalchemy import inspect, case, null
 from sqlalchemy.orm.properties import ColumnProperty
+from sqlalchemy import func
```
kevin-pierce: nit:
```suggestion
from sqlalchemy import inspect, case, null, func
from sqlalchemy.orm.properties import ColumnProperty
```

backend/app/services/implementations/user_service.py
```diff
@@ -113,14 +113,20 @@ def get_auth_id_by_user_id(self, user_id):
...
-                    User.query.limit(results_per_page)
+                    User.query
+                    .order_by(User.last_modified.desc())
+                    .limit(results_per_page)
                     .offset((page_number - 1) * results_per_page)
                     .all()
-                )
-
+                    )
+
```
kevin-pierce: nit: Undo this formatting for consistency w the other queries

kevin-pierce (approved):
LGTM 👍

backend/app/resources/auth_dto.py
```diff
@@ -13,6 +13,7 @@ def __init__(
         email,
         role,
         user_status,
+        last_modified
```
connor-bechthold: Are the changes in these dto files still required? I thought we could remove them but I may be wrong

backend/migrations/versions/698e5724baae_last_modified_column.py
```diff
@@ -0,0 +1,32 @@
...
+revision = '698e5724baae'
+down_revision = '0ea2257f1dc6'
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
+    # ### commands auto generated by Alembic - please adjust! ###
+    with op.batch_alter_table('users', schema=None) as batch_op:
+        batch_op.add_column(sa.Column('last_modified', sa.DateTime(), nullable=True))
```
connor-bechthold: so one thing i noticed is when you run the initial invite user script it doesn't work because this doesn't actually have a default - I believe if you add these:
```
server_default=text('CURRENT_TIMESTAMP'), onupdate=text('CURRENT_TIMESTAMP')
```

it should solve this issue.

Also, make sure to set this as `nullable=False`

backend/app/models/user.py
```diff
@@ -19,6 +19,9 @@ class User(db.Model):
         nullable=False,
     )
     email = db.Column(db.String, nullable=False)
+    last_modified = db.deferred(
+        db.Column(db.DateTime, default=func.now(), onupdate=func.now())
```
connor-bechthold: After I researched we should be using `server_default` instead of `default`. Basically this takes the timestamp from the DB instead of the server, which is more reliable.

Also, ensure this is non-nullable: `nullable=False`

connor-bechthold (changes requested):
Left some comments - if any of these don't make sense let me know!!

connor-bechthold (approved):
muy bien

PR #200, opened 2023-11-12
Unique Resident IDs

kevin-pierce:
## Notion task link

[Unique Resident ID](https://www.notion.so/uwblueprintexecs/2792c3504d744d2b90ff967c32d00961?v=d33629815ecf477ab4e1ff4e33774c32&p=aaf516a82f2542b393b676a90b8c31c9&pm=s)

## Implementation description
* Added new response type for editing / adding functions of resident client
* Added error toast on error response on FE
* Added helper to type narrow on error response type for result
* Added logic to resident routes to call **get_residents** prior to editing / updating residents

### Create Resident
1. Create a resident
2. Attempt to create another resident with the same initials + room number (Expect: error toast, no change to residents table state)
3. Attempt to create another resident with different initials + same room number OR same initials + different room number OR different initials + different room number (Expect: success toast, new resident added to list of residents)

### Edit Resident
1. Create two resident unique residents
2. Attempt to edit one of the residents such that the initials + room number match those of the other resident (Expect: error toast, no change to residents table state)
3. Attempt to edit the same resident with no changes (Expect: success toast)
4. Attempt to edit a resident and set their initials + room number to be another non-existent combination (Expect: success toast)

* Optimizations around how we can avoid making the duplicate check calls on every addition + update of a resident (I was thinking of something involving the frontend and how we can check what fields were modified ; accordingly, we send back a flag / some status that indicates that we **do / do NOT** need to perform this check)

## Checklist


backend/app/rest/residents_routes.py
```diff
@@ -21,6 +21,16 @@ def add_resident():
             400,
         )
 
+    # Check for the existence of a resident prior to adding them
+    fmt_resident_id = resident.get("initial") + str(resident.get("room_num"))
+    try:
+        res = residents_service.get_residents(False, 1, 10, fmt_resident_id)
+        if len(res["residents"]) > 0:
```
kevin-pierce, line 28: When adding a new resident, we simply want to check for a result length greater than 0 (if there's even one other resident that has the same ID, then this is a duplicate resident => reject)

backend/app/rest/residents_routes.py
```diff
@@ -41,6 +51,17 @@ def update_resident(resident_id):
             jsonify({"date_left_error": "date_left cannot be less than date_joined"}),
             400,
         )
+    
+    # Check for the existence of a resident prior to adding them
+    fmt_resident_id = updated_resident.get("initial") + str(updated_resident.get("room_num"))
+    try:
+        res = residents_service.get_residents(False, 1, 10, fmt_resident_id)
+        if len(res["residents"]) == 1 and res["residents"][0]["id"] != resident_id:
```
kevin-pierce, line 59: When editing a resident, there are two cases for which the result of **get_residents** could be `>=1` in size:

1. There was no change to the resident ID (i.e. a resident with ID AA123 has their building num changed ; ID remains AA123 and so when we get residents with the ID `fmt_resident_id = AA123`, then we get 1)

2. We changed another resident to have a pre-existing resident ID (there are residents AA123 and AA321 ; resident AA123 has their room number changed to 321, resulting in two duplicate entries ; thus when we fetch the residents with the ID AA321, we end up with a result of 1)

We want case #2 to be marked as INVALID but case #1 to be VALID. As such, our checking logic must check for:
- If there's already a result being returned
- If the result has the same database ID as us (if it's the same, then we're editing the same record, meaning this change is valid ; if the database IDs are **different**, then this means that there is a _different_ resident with the same formatted ID as us)

frontend/src/APIClients/ResidentAPIClient.ts
```diff
@@ -73,6 +74,13 @@ const createResident = async ({
     );
     return true;
   } catch (error) {
+    const axiosErr = (error as any) as AxiosError;
+
+    if (axiosErr.response && axiosErr.response.status === 409) {
+      return {
+        errMessage: "Resident with the specified user ID already exists."
```
kevin-pierce, line 81: Here, we're propagating the error message displayed in the toast through the API client

frontend/src/components/forms/CreateResident.tsx
```diff
@@ -61,17 +63,30 @@ const CreateResident = ({
...
+        "Error creating resident",
+        res.errMessage,
+        "error"
+      )
+    }
+    else if (res) {
+      getRecords(1);
+      countResidents();
+      setUserPageNum(1);
+      setShowAlert(true)
```
kevin-pierce, line 84: We move the call to here since the showAlert state controls the visibility of a `success` alert ; we only want the success state to be displayed if we get back a boolean result with value `true`

braydonwang (approved):
LGTM! I tested creating and editing residents on the frontend and everything works as expected, and I didn't find any real problems related to the code

connor-bechthold (approved):
THE EXPLANATIONS ARE 💯 , one thing I think is worth changing is keeping the modal open (and not clearing whatever was entered) when a duplicate resident is entered

Also I'm thinking since we have a few more error types/messages now it might be worth having them in their own shared files:
- Can have a file `ErrorTypes.ts` (under `/types`) which can hold the `ErrorResponse` here and the one defined in `https://github.com/uwblueprint/supportive-housing/pull/202/files`
- Under `helpers` we have a file called `errors.ts` which can hold everything in `residentError.ts` and `authError.ts`

Just so it makes things more reusable if we ever need to use the functionality in the future.

PR #201, opened 2023-11-16
Cascading Delete Tags

danielk1345:
## Notion task link

[Task Name](https://www.notion.so/uwblueprintexecs/Cascading-Tags-58fb5fab828147748139026c4a4d81dc?pvs=4)

## Implementation description
* Implemented the cascading delete so that it deletes all log record references with the associated tag and replaces that status field in the tags database

## Steps to test
1. Migrate your database and upgrade
2. Create sample tags in your tags database. Ex: INSERT into tags (name, status) VALUES ('tagA', 'Active');
3. Then, in the browser, create log records that have tags associated with them. Example body: 
`{
  "employee_id": 7,
  "resident_id": 2,
  "datetime": "2023-11-12T01:56:01.891Z",
  "flagged": false,
  "note": "lalala",
  "tags": ["tagA", "tagB", "tagC"],
  "building_id": 1
}
`
4. Then, try deleting a tag that is associated with the log record. Verify in your log_record_tag junction table that all entries with that tag_id have now been deleted.
5. Verify that the status column no longer exists in your database

* 

## Checklist


backend/app/models/tags.py
```diff
@@ -1,5 +1,6 @@
 from sqlalchemy import inspect
 from sqlalchemy.orm.properties import ColumnProperty
+from sqlalchemy.orm import backref
```
braydonwang: This import is unused, I think it can be removed

backend/migrations/versions/117790caec65_.py
```diff
@@ -0,0 +1,32 @@
...
+
+def upgrade():
+    # ### commands auto generated by Alembic - please adjust! ###
+    with op.batch_alter_table('tags', schema=None) as batch_op:
+        batch_op.drop_column('status')
+
+    # ### end Alembic commands ###
+
+
+def downgrade():
```
braydonwang, line 27: Running "flask db downgrade" fails and complains that the "status" column contains null values. I think it's safe to set all the tags in the db to "Active" when downgrading, since they wouldn't exist in the db if they were deleted.

braydonwang (approved):
Other than the two comments I left, everything else looks good! I tested it through Postman and everything works as expected. I'll approve this PR for now since the problems I found aren't too big

phamkelly17:
Screen Shot 2023-11-22 at 9 06 58 PM

not sure if somethings wrong on my end, but i'm not able to delete tags since theyre referenced in the `log_record_tag` table

phamkelly17:
in this file `supportive-housing/backend/app/models/log_record_tags.py`, would you need something like this:

```
    log_record_tag_id = db.Column(db.Integer, primary_key=True, nullable=False)
    log_record_id = db.Column(
        db.Integer, db.ForeignKey("log_records.log_id"), nullable=False
    )
    tag_id = db.Column(db.Integer, db.ForeignKey("tags.tag_id"), nullable=False, cascade="all, delete")
```

idk at all but just a thought

backend/migrations/versions/117790caec65_.py
connor-bechthold: I think this is okay - any existing tags created after this is merged shouldn't have the `status` column anyways so that should solve things

connor-bechthold:
> in this file `supportive-housing/backend/app/models/log_record_tags.py`, would you need something like this:
> 
> ```
>     log_record_tag_id = db.Column(db.Integer, primary_key=True, nullable=False)
>     log_record_id = db.Column(
>         db.Integer, db.ForeignKey("log_records.log_id"), nullable=False
>     )
>     tag_id = db.Column(db.Integer, db.ForeignKey("tags.tag_id"), nullable=False, cascade="all, delete")
> ```
> 
> idk at all but just a thought

so if you try and use the DELETE route (`/tags/{id}`) this works fine, however you're right that we don't really have cascading set up properly here. we should still be able to go and delete a tag manually and not have any foreign key constraint error. kelly is right about where the change is however I think we tried this before and it didn't work. i looked a little bit more and tested this out and it didn't complain with this (after creating a migration and upgrading)
```
tag_id = db.Column(db.Integer, db.ForeignKey("tags.tag_id", ondelete='CASCADE'), nullable=False)
```

all this does is trigger a cascading delete whenever that foreign key is deleted as well

connor-bechthold (approved):
awesome LGTM!!!

PR #202, opened 2023-11-16
Sign-up Page Functionality

kevin-pierce:
## Notion task link

[Signup Page Functionality](https://www.notion.so/uwblueprintexecs/2792c3504d744d2b90ff967c32d00961?v=d33629815ecf477ab4e1ff4e33774c32&p=dbd76d84f8284efc952413d2627a9105&pm=s)

## Implementation description
* Added two exceptions ; UserNotInvitedException and EmailAlreadyInUseException with corresponding error messages
* Added 3 error states for email (invalid email, user not invited, email already in use)
* Cleaned up dead code (unused `Signup.tsx` file)
* Restyled `Signup.tsx` form to use Chakra built-in components

### Invalid email
1. Attempt registration flow with invalid email (for example, `testatshouldnotwork.com`) 
2. Verify that we see an "Invalid email" error appear below the email entry field
3. Verify that when we FIRST enter an invalid email, that we DO NOT see the error appear ; we should only see this type of error appear AFTER we have clicked the signup button 

### User Not Invited
1. Attempt to register a user with an email that has not been invited (doesn't exist in your database)
2. Verify that we see an "Email not invited" error appear below the email entry field

### Email Already in Use
1. Attempt to register a user with an email already associated with an `Active` account 
2. Verify that we see an "Email already in use" error appear below the email entry field

### General
- Verify that the sign up button remains disabled when any of the fields are blank
- Verify that the sign up button remains disabled when an invalid email is present
- Verify that successful registration flow operates as normal

* 

## Checklist


backend/app/rest/auth_routes.py
```diff
@@ -3,6 +3,7 @@
     InvalidPasswordException,
```
braydonwang, line 3: Not sure if this needs to be addressed in this PR, but I was trying to sign up with an account that is "Invited" but clicking "Create Account" didn't do anything. I looked in the networks tab and found that the register endpoint failed with the error: "Invalid password string. Password must be a string at least 6 characters long." This could be unintuitive for the users since it didn't show an error box so they don't know that this password character limit exists.

braydonwang (approved):
Besides the comment I left, everything else looks good and works well!

backend/app/rest/auth_routes.py
kevin-pierce: Resolved ; I approached this by offloading the password verification to the frontend (by simply doing a length check to verify the password is >= 6 characters long). This check is performed both in the onClick handler for createAccount and within the handlePasswordChange function (which functions the same way as the valid email error does, in which we only check for the validity of these fields _after_ the create account button has been clicked)

frontend/src/components/forms/Signup.tsx
```diff
@@ -32,38 +43,98 @@ const Signup = ({
...
-        const { requiresTwoFa, authUser } = registerResponse;
-        if (requiresTwoFa) {
-          setToggle(!toggle);
-        } else {
-          localStorage.setItem(
-            AUTHENTICATED_USER_KEY,
-            JSON.stringify(authUser),
-          );
-          setAuthenticatedUser(authUser);
+    if (isInvited !== "Not Invited") {
```
kevin-pierce, line 101: The above check would trigger as long as some string was being returned here (which this function always did, even in the error case) hence we just changed the logic around

frontend/src/helper/authError.ts
```diff
@@ -0,0 +1,15 @@
+import { AxiosError } from "axios";
+import { AuthTokenResponse, ErrorResponse, AuthFlow } from "../types/AuthTypes";
+
+export const getAuthErrMessage = (axiosErrRes: AxiosError["response"], flow: AuthFlow): string => {
+  if (axiosErrRes && axiosErrRes.data && axiosErrRes.data.error) {
+    return axiosErrRes.data.error;
+  }
+  return `Error ${flow === 'LOGIN' ? "logging in" : "signing up"}. Please try again later.`;
+}
```
kevin-pierce, line 9: Takes in the auth flow (either 'LOGIN' or 'SIGNUP') from which we're accessing this method, as to consolidate the two functiosn we had before into one

frontend/src/components/forms/Login.tsx
```diff
@@ -26,12 +27,6 @@ type CredentialsProps = {
   setToggle: (toggle: boolean) => void;
 };
 
-const isLoginErrorResponse = (
-  res: AuthTokenResponse | ErrorResponse,
-): res is ErrorResponse => {
-  return res !== null && "errCode" in res;
-};
-
```
kevin-pierce, line 34: Added a more comprehensive version of this function `isAuthErrorResponse`

frontend/src/components/forms/Signup.tsx
danielk1345: Could you explain what you mean by this? Do you mean that it would show not invited even if an invalid string was passed in?

connor-bechthold (approved):
muy bien

danielk1345 (approved):
Looks good, and it works as expected. Only thing that I kind of struggled to understand was one of your comments

frontend/src/components/forms/Signup.tsx
kevin-pierce: Basically, before the condition was simply `if (isInvited)`. However, the function call was returning (even in the error case) the string `Not Invited`. This meant that, even when we were _supposed_ to be erroring out, we were interpreting the API call as successful since `if (isInvited)` would be true if the string was not empty

PR #203, opened 2023-11-19
Connecting tags when creating, filtering and editing log record

braydonwang:
## Notion task link

[Task](https://www.notion.so/uwblueprintexecs/Connecting-Tags-To-Log-Records-c0ba6affff9646068320177bbef67c46)

## Implementation description
* Enable multi-select tags input when the user creates, edits or filters a log record
* Make GET request to backend to get all ACTIVE tags
* Change backend logic to use tag id (rather than tag name) when creating or filtering a log record

## Steps to test
1. At the moment, we don't have any way to create a tag from the frontend or backend, so you'll need to manually insert it into the db
```
insert into tags (tag_id, name, status) values (1, 'Test Tag', 'Active');
```
2. On the frontend, when creating a log record, click the "Tags" input box and check that the dropdown displays all tags that are currently in the tags table in the db
3. Add a couple of log records with: a) No tags, b) One tag, c) Multiple tags (make sure that you get a success response on the frontend and check that the log_record_tag table has also been correctly updated)
4. Try editing log records and updating their tags section: a) Remove all tags, b) Add new tags c) Don't change tags (similar to create, make sure this gives successful response and updates log_record_tag table accordingly)
5. Filter by a single tag or multiple tags and ensure that all log records with ALL of the specified tags are being displayed (also check that the count of log records is correct)

* Let me know if I should split this into multiple PRs (backend and frontend) to make it easier to review

## Checklist
- [X] My PR name is descriptive and in imperative tense
- [X] My commit messages are descriptive and in imperative tense. My commits are atomic and trivial commits are squashed or fixup'd into non-trivial commits
- [X] I have run the appropriate linter(s)
- [X] I have requested a review from the PL, as well as other devs who have background knowledge on this PR or who will be building on top of this PR
- [X] If I have made API changes, I have updated the [REST API Docs](https://www.notion.so/uwblueprintexecs/REST-Endpoints-05ce60312bb943439dfda42bb1318536)
- [X] IF I have made changes to the db/models, I have updated the [Data Models Page](https://www.notion.so/uwblueprintexecs/Data-Models-760f8aa06b244eb0842c079ad77987b0)
- [X] I have documented this PR in the [Production Release Page](https://www.notion.so/uwblueprintexecs/fe28a97bfa5648d3bc3795b32b694acd?v=5565cfb35dcc45bb84f6528cb09517cd)
- [X] I have updated other Docs as needed


phamkelly17 (approved):
works perfectly, lgtm

connor-bechthold (approved):
PERFECT 👨‍🍳  thanks again for changing the log record service to use tag ID instead of name 🙏 I'll make a ticket for this so it can be tracked

Only think to mention is hold of merging until https://github.com/uwblueprint/supportive-housing/pull/201 is through - this removes the `status` field with the cascading so you won't need to reference it in this PR anymore (filtering by only `Active` tags was def the right thinking tho)

PR #208, opened 2023-11-25
Note, Tags, and Residents Column Formatting

owen-sellner:
## Notion task link

[Notes Column Limitation](https://www.notion.so/uwblueprintexecs/Notes-Column-Limitation-d54b79f5b52b403fba198ca1f47c0b65)
[Tags Column](https://www.notion.so/uwblueprintexecs/Tags-Column-1fb05a3bb1b64d1c9be32349ed82c15a)
[Residents Column Limitation](https://www.notion.so/uwblueprintexecs/Residents-Column-Limitation-eace22b5f3f746f7a669d47a49ebc34c)

## Implementation description
* Added formatting functions to `LogRecordsTable.tsx`
* Added extra null check in `EditLog.tsx` because of a TypeScript/eslint issue
* Fixed bug with table header where the options button overlapped the heading by adding a z-index value to `tableStyles.tsx`

## Steps to test
1. Create new log records with 1 resident, 2 residents, 3 residents, and 4 residents
2. Create new log records with a note less than 150 characters and one greater than 150 characters
3. Create new log records with 1 tag, 2 tags, 3 tags, and 4 tags
4. Check that formatting for note, tags, and residents is correct

* How many characters should be shown for the note?

## Screenshot
image

## Checklist

### DIFF
--- a/frontend/src/components/forms/EditLog.tsx
+++ b/frontend/src/components/forms/EditLog.tsx
@@ -218,7 +218,7 @@ const EditLog = ({
     );
     setBuildingId(logRecord.building.id);
     const residentIds = residentOptions.filter(
-      (item) => logRecord.residents.includes(item.label),
+      (item) => logRecord.residents && logRecord.residents.includes(item.label),
     ).map((item) => item.value);
     setResidents(residentIds);
     setTags(logRecord.tags);
@@ -369,7 +369,7 @@ const EditLog = ({
                       placeholder="Select Residents"
                       onChange={handleResidentsChange}
                       defaultValue={residentOptions.filter(
-                        (item) => logRecord.residents.includes(item.label),
+                        (item) => logRecord.residents && logRecord.residents.includes(item.label),
                       )}
                       styles={selectStyle}
                     />
--- a/frontend/src/components/pages/HomePage/LogRecordsTable.tsx
+++ b/frontend/src/components/pages/HomePage/LogRecordsTable.tsx
@@ -85,6 +85,16 @@ const LogRecordsTable = ({
     }));
   };
 
+  const formatNote = (note: string) => {
+    if (note && note.length > 0) {
+      if (note.length > 150) {
+        return note.substring(0, 150).concat("...");
+      }
+      return note;
+    }
+    return "";
+  };
+
   const formatList = (strArr: string[]) => {
     const strLength = strArr?.length;
     if (strLength === 1) {
@@ -191,7 +201,7 @@ const LogRecordsTable = ({
                         {formatList(record.residents)}
                       </Td>
                       <Td whiteSpace="normal" width="65%">
-                        {record.note}
+                        {formatNote(record.note)}
                       </Td>
                       <Td width="5%">{`${record.employee.firstName} ${record.employee.lastName}`}</Td>
                       <Td width="5%">
--- a/frontend/src/components/pages/HomePage/LogRecordsTable.tsx
+++ b/frontend/src/components/pages/HomePage/LogRecordsTable.tsx
@@ -45,6 +45,28 @@ const DELETE_CONFIRMATION_HEADER = "Delete Log Record";
 const DELETE_CONFIRMATION_MESSAGE =
   "Are you sure you want to delete this log record? Deleting a log record will permanently remove it from your system.";
 
+const formatNote = (note: string) => {
+  const NOTE_LIMIT = 150;
+  if (note.length > NOTE_LIMIT) {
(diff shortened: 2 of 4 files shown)


frontend/src/components/pages/HomePage/LogRecordsTable.tsx
```diff
@@ -85,6 +85,31 @@ const LogRecordsTable = ({
     }));
   };
 
+  const formatNote = (note: string) => {
+    const noteLimit = 150;
```
kevin-pierce: nit: Move this up higher in the file (reduces memory of function invocation) + renaming it to adhere to other constant names

```suggestion
    const NOTE_LIMIT = 150;
```

owen-sellner: I will make the change but just for my own learning how does moving the function reduce memory usage?

connor-bechthold: @owen-sellner basically everytime you invoke this function you're declaring this as a new variable - since it never changes you can put it as a constant at the top of the file so it's only declared once

frontend/src/components/forms/EditLog.tsx
```diff
@@ -218,7 +218,7 @@ const EditLog = ({
     );
     setBuildingId(logRecord.building.id);
     const residentIds = residentOptions.filter(
-      (item) => logRecord.residents.includes(item.label),
+      (item) => logRecord.residents && logRecord.residents.includes(item.label),
```
connor-bechthold, line 221: hmm curious as to why this was required since we should be guaranteed to have residents on a log record all the time

frontend/src/components/pages/HomePage/LogRecordsTable.tsx
```diff
@@ -85,6 +85,31 @@ const LogRecordsTable = ({
...
+    if (note && note.length > 0) {
```
connor-bechthold: would we have to check for a note here? I don't think there'd ever be a case when a note is empty

frontend/src/components/pages/HomePage/LogRecordsTable.tsx
```diff
@@ -85,6 +85,31 @@ const LogRecordsTable = ({
...
+      if (note.length > noteLimit) {
+        return note.substring(0, noteLimit).concat("...");
+      }
+      return note;
+    }
+    return "";
+  };
+
+  const formatList = (strArr: string[]) => {
+    const strLength = strArr?.length;
```
connor-bechthold: Should be able to get rid of the `?` operator - if there's no results, you'll be passed in an empty array which has a length of 0.

frontend/src/components/pages/HomePage/LogRecordsTable.tsx
```diff
@@ -173,17 +199,20 @@ const LogRecordsTable = ({
                       <Td width="5%">{date}</Td>
                       <Td width="5%">{time}</Td>
                       <Td whiteSpace="normal" width="5%">
-                        {record.residents?.join("\n")}
+                        {formatList(record.residents)}
                       </Td>
-                      <Td whiteSpace="normal" width="70%">
-                        {record.note}
+                      <Td whiteSpace="normal" width="65%">
```
connor-bechthold: So ideally the width should remain the same - to get more horizontal space I think we can get away with only displaying the first name of the user in the `Employee` and `Attn To` fields a couple lines below.

connor-bechthold (changes requested):
So I noticed the `zIndex` change may not have worked as intended:
Screenshot 2023-12-02 at 8 38 52 PM

Left a couple other small comments but otherwise really well done!!

frontend/src/components/forms/EditLog.tsx
kevin-pierce: Saw this pop up on Kelly's PR as well ; could it be a linting / warning resolution thing?

owen-sellner: Yeah, I am pretty sure it is a lint issue but it wouldn't compose until I added this

frontend/src/components/pages/HomePage/LogRecordsTable.tsx
owen-sellner: It was causing a typescript issue until I added the question mark

phamkelly17 (approved):
there's a bug though l.o.l

connor-bechthold (approved):
beauty

PR #214, opened 2023-12-02
Spinner

phamkelly17:
## Notion task link
n/a

## Implementation description
* Before, the header of the table would appear on the screen before the table data was available. Table data would appear at a delay.
* adding a spinner if we're in the process of fetching table data
* adding a "no results found" message if no results are founf

## Steps to test
1. go to home page
2. check if the spinner is there and looks good while the table is loading
3. make sure the table loads
4. filter so that there are 0 results and check if it gives a message that says "No results found."
5. repeat for resident and employee directory pages (step 4 not necessary)

## Checklist


frontend/src/components/pages/AdminControls/EmployeeDirectory.tsx
```diff
@@ -71,23 +71,34 @@ const EmployeeDirectoryPage = (): React.ReactElement => {
...
-          setResultsPerPage={setResultsPerPage}
-          getRecords={getUsers}
-        />
+        {numUsers < 0 ? (
+          <Spinner
+            thickness="4px"
+            speed="0.65s"
+            emptyColor="gray.200"
+            size="xl"
+          />
```
phamkelly17, line 85: there is no way to put these props in the default props section in the SpinnerStyles.tsx file. it just won't work. trust me, me and gpt have tried

frontend/src/components/pages/AdminControls/EmployeeDirectory.tsx
```diff
@@ -71,23 +71,34 @@ const EmployeeDirectoryPage = (): React.ReactElement => {
...
+        ) : (
```
phamkelly17, line 86: didn't think checking for "No results found" was necessary because if you're using the website, that means you are a user and there is at least 1 employee. will there be employee filtering later or am i good to leave as is

connor-bechthold: you're good

frontend/src/components/forms/EditLog.tsx
```diff
@@ -375,7 +375,7 @@ const EditLog = ({
                       placeholder="Select Residents"
                       onChange={handleResidentsChange}
                       defaultValue={residentOptions.filter(
-                        (item) => logRecord.residents.includes(item.label),
+                        (item) => logRecord.residents && logRecord.residents.includes(item.label),
```
connor-bechthold, line 378: is there some sort of error that appears that required this to be changed? residents should never be null

frontend/src/components/pages/AdminControls/EmployeeDirectory.tsx
connor-bechthold: skill issue

connor-bechthold (changes requested):
this looks great however it would be nice if this spinner could also show when something is being added/edited/deleted (since it triggers a refetch of the table)

what you can do is have a loading state that's initially set to false. everytime you refetch new records (i.e. call `getLogRecords()`), you can set the loading state to `True` before the call, and then `False` after the call

ex. 
```
setIsLoading(True)
const data = await LogRecordAPIClient.filterLogRecords({
      buildingId: buildingIds,
      employeeId: employeeIds,
      attnTo: attentionToIds,
      dateRange: dateRange[0] === "" && dateRange[1] === "" ? [] : dateRange,
      residents: residentsIds,
      tags: tagsValues,
      flagged,
      resultsPerPage,
      pageNumber,
    });
setIsLoading(False)
```

that way the spinner can show on inital page renders AND adds/edits/deletes. the rest of the logic you implemented should be able to stay the same

frontend/src/components/pages/AdminControls/EmployeeDirectory.tsx
kevin-pierce: So as Connor mentioned, the best way to implement this is usually using a state variable that simply keeps track of the loading state as a boolean ; we then would conditionally render the spinner only if "loading" is taking place.

Here's a working snippet

```
// Initially set to false ; this way by DEFAULT we display the spinner
const [loadedUsers, setLoadedUsers] = useState(false)

// At the "onMount" useEffect (the one w no dependencies), we'd want to set the loaded users state to true
// The only reason I don't want to set it in the other useEffect as well is because from what I understand, we don't want to display the spinner for every batched fetch of users
  useEffect(() => {
    countUsers();
    setLoadedUsers(true)
  }, []);

// Then below in the JSX:
{!loadedUsers ? <Spinnner> : <other stuff>}
```

frontend/src/components/forms/EditLog.tsx
phamkelly17: i don't recall changing this file in this PR, but yeah there is some error that requires this to be changed. i believe owen made a pr for it

kevin-pierce (approved):
LGTM 👍

connor-bechthold (approved):
lgtm!!!

